PR #291, opened 2015-05-25
Fix a typo

mootoh:
Just a little bit.

### DIFF
--- a/doc/build_mozc_in_docker.md
+++ b/doc/build_mozc_in_docker.md
@@ -121,7 +121,7 @@ python build_mozc.py build -c {Release, Debug} [gyp_path_1.gyp:gyp_target_name1]
 
 ## Android specific topics
 
-### Aplication package name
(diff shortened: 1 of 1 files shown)


yukawa:
Thank you for PR. Unfortunatelly we are not accepting patches right now as explained in [CONTRIBUTING.md](https://github.com/google/mozc/blob/master/CONTRIBUTING.md).  Technically we can, but doing that means that Mozc is no longer exclusively copyrighted by Google even if you sign our Contributor License Agreement, and due to our internal third party code policy, which also means that we need to change the code location of internal version of Mozc in our interenal code repository from the current location to third_party/mozc.  This is non-trivial task because many of internal projects directly or indirectly depend on Mozc.

Regarding the typo you found, I'll ask someone in Mozc team to fix it to avoid copyting issues.  Thank you again for letting us know about that.

PR #382, opened 2016-09-06
Fix typo

ikegami-yukino:
`~/wrk/mozc/src/out_mac/Release` to `~/work/mozc/src/out_mac/Release`

PR #397, opened 2016-11-23
fix typo in document

lv7777:
`Release`` or `Debug` -> `Release` or `Debug`.


yukawa:
Thank you for the patch, but as we explained in #382, we are not currently accepting any external contributions since it would require us to move more than 2,000 file in our internal repository.  Good news is that the typo is already addressed by 280e38fe3d9db4df52f0713acf2ca65898cd697a so we can just close this thread you don't need to sign our CLA (Contributor License Agreement) anymore.

Thanks again for your patch and we sincerely apologize for not being able to accept your patch.

PR #478, opened 2020-05-07
Fix input mode selection status

roberthoffmann:
This fixes the problem described in
https://github.com/ibus/ibus/issues/2026


roberthoffmann:
@googlebot I signed it!

hiroyuki-komatsu:
Thank you very much for your pull request, and I'm sorry for my late response.

The file location `src/unix/ibus` was one of directories that we did not accept PRs.
We finally added  this directory to the directories we can accept PRs.

* https://github.com/google/mozc/commit/f133df0fd92c82945259bb588ea227d6738b1d06
* https://github.com/google/mozc/issues/567

It seems the issue has been already addressed in a similar approach.
Best regards,

PR #522, opened 2021-09-14
Add entries to aux_dictionary.tsv and oss.tsv

utuhiro78:
I haven't tested oss.tsv. It needs Bazel.


utuhiro78:
@googlebot I signed it!

PR #549, opened 2022-02-25
Use apt-get instead of apt in Dockerfile

akihikodaki:
Ubuntu Manpage: apt - command-line interface
SCRIPT USAGE AND DIFFERENCES FROM OTHER APT TOOLS
https://manpages.ubuntu.com/manpages/focal/en/man8/apt.8.html#script%20usage%20and%20differences%20from%20other%20apt%20tools
> The apt(8) commandline is designed as an end-user tool and it may
> change behavior between versions. While it tries not to break backward
> compatibility this is not guaranteed either if a change seems
> beneficial for interactive use.
>
> All features of apt(8) are available in dedicated APT tools like
> apt-get(8) and apt-cache(8) as well.  apt(8) just changes the default
> value of some options (see apt.conf(5) and specifically the Binary
> scope). So you should prefer using these commands (potentially with
> some additional options enabled) in your scripts as they keep backward
> compatibility as much as possible.

### DIFF
--- a/docker/ubuntu20.04/Dockerfile
+++ b/docker/ubuntu20.04/Dockerfile
@@ -34,20 +34,20 @@ ENV DEBIAN_FRONTEND noninteractive
 # Package installation
 RUN apt-get update
 ## Common packages for linux build environment
-RUN apt install -y clang libc++-dev libc++abi-dev python python-six python3-six pkg-config git curl bzip2 unzip make ninja-build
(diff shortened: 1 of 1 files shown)


shitamo:
>  WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
I got a warning when I made apt

hiroyuki-komatsu:
Thank you for the patch and the confirmation of the warning.

Issue #567, opened 2022-04-14
GYP build will stop supporting IBus client.


hiroyuki-komatsu:
We are going to have following changes

* Gtk2 based candidate window for IBus will be removed.
  + Qt based candidate window is the alternative.
* GYP build will stop supporting the build of IBus client.
  + Bazel build is the alternative. 
* Then, pull requests for IBus client and Qt based candidate window will be acceptable.
 
We need to stop supporting GYP build to accept pull requests because of the limitation of the upstream.
We believe accepting pull requests is more beneficial than keeping the GYP build under the maintenance mode.

We will start this change on 2022-05 at the earliest.

stalkerg:
Sorry, but Bazel it's pain in the ass for Linux distro maintainers and source-based distro users. Just to build a C++ project for i18n we should somehow install JVM and Bazel - it's very heavy requirements. 
We should consider adding Meson build and add fcitx5 support. 
I know - Google can do what they want, but this is what is needed for the community. 
Google is not too poor a company to not support the whole Linux ecosystem in the right way. 
Can we start any negotiations about that?

hiroyuki-komatsu:
Hi, thank you for the feedback.

Because GYP is not updated well, we need to make patches to keep supporting builds of Mozc.
Bazel is the default tool used by the upstream of the Mozc project and the number of people keep improving Bazel as full time commitments. Bazel improves our productivity.

So we would like you to encourage us to focus our engineering resource on the core functionalities of Japanese input rather than build systems.

Thank you,

stalkerg:
> Because GYP is not updated well, we need to make patches to keep supporting builds of Mozc.

yes, GYP is outdated but on Linux desktop, Meson and CMake are dominated now and very easily integrated. 

> Bazel is the default tool used by the upstream 

I suppose it's a mistake. 

> So we would like you to encourage us to focus our engineering resource on the core functionalities of Japanese input rather than build systems.

because of this, I can't use mozc at all!!!! Please, read my original message again, I asked and point to a completely different issue.

akihikodaki:
In my opinion, having a JDK dependency at build time can be tolerated. It is understandable that the upstream uses the best tool for core productivity. As far as I know, having JDK as a build time dependency is not a problem for APT or RPM packages anyway. 

However, I have another concern: `src/config.bzl` has hardcoded configurations. Particularly, Fedora places Qt to a different path. I have just skimmed the code and Bazel documentation but it looks like there is no straightforward method to make it configurable.

[...]

PR #637, opened 2022-12-02
Use GNOME Shell's panel instead of mozc_renderer under GNOME.

AlynxZhou:
GNOME Shell provides an integrated candidate panel for ibus, which looks better than mozc_renderer. And because the panel is part of compositor itself, it works better than mozc_renderer under Wayland because Wayland clients are not allowed to set their position.

**Pull requests to the Mozc project are limited to the specific directories.**

Files and directories we may accept pull requests:
* files in the [top directory](https://github.com/google/mozc/tree/master/)
* [.github/](https://github.com/google/mozc/tree/master/.github/)
* [docker/](https://github.com/google/mozc/tree/master/docker/)
* [docs/](https://github.com/google/mozc/tree/master/docs/)
* [src/.bazelrc](https://github.com/google/mozc/tree/master/src/.bazelrc)
* [src/BUILD.(library).bazel](https://github.com/google/mozc/tree/master/src/)
* [src/data/oss/](https://github.com/google/mozc/tree/master/src/data/oss/)
* [src/data/test/quality_regression_test/](https://github.com/google/mozc/tree/master/src/data/test/quality_regression_test/)
* [src/renderer/qt/](https://github.com/google/mozc/tree/master/src/renderer/qt/)
* [src/unix/emacs/mozc.el](https://github.com/google/mozc/tree/master/src/unix/emacs/mozc.el)
* [src/WORKSPACE.bazel](https://github.com/google/mozc/tree/master/src/WORKSPAE.bazel)

Although Google company policy certainly allows Mozc team to accept pull
requests, to do so Mozc team needs to move all Mozc source files into
`third_party` directory in the Google internal source repository [1].
Doing that without breaking any Google internal project that depends on
Mozc source code requires non-trivial amount of time and engineering
resources that Mozc team cannot afford right now.

Mozc team continues to seek opportunities to address this limitation,
but we are still not ready to accept any pull request due to the above
reason.

[1]: [Open Source at Google - Linuxcon 2016](http://events.linuxfoundation.org/sites/events/files/slides/OSS_at_Google.pdf#page=30)
> ### License Compliance
> - We store all external open source code in a third_party hierarchy,
> along with the licenses for each project. We only allow the use of OSS
> under licenses we can comply with.
> - Use of external open source is not allowed unless it is put in that
> third_party tree.
> - This makes it easier to ensure we are only using software with
licenses that we can abide.
> - This also allows us to generate a list of all licenses used by any
project we build when they are released externally.

### DIFF
--- a/src/unix/ibus/mozc_engine.cc
+++ b/src/unix/ibus/mozc_engine.cc
@@ -251,6 +251,12 @@ CandidateWindowHandlerInterface *createGtkCandidateWindowHandler(
   }
 #endif  // !ENABLE_QT_RENDERER
 
+  if (GetEnv("XDG_CURRENT_DESKTOP") == "GNOME") {
+    // In GNOME, ibus can use candidate panel provided by GNOME Shell, which
+    // works better than mozc_renderer.
+    return nullptr;
+  }
+
   auto *handler = new GtkCandidateWindowHandler(renderer_client);
   handler->RegisterGSettingsObserver();
   return handler;


AlynxZhou:
Wait, is this meaning you guys does not accept patchs for `ibus-mozc`? However the current Qt renderer is **totally unusable** under GNOME Wayland because I have two monitors and it's always on the top left corner of my primary monitor.

I don't care about company shits, I just want to fix my desktop experience, so if you guys have trouble on accept my PR, just take this as an issue and directly copy my code and commit by yourself please. If it's acceptable, add my name to commit message, if your company does not allow you to do that, it's OK, just copy those code to fix it.

AlynxZhou:
This is what the Qt renderer looks like under GNOME Wayland, totally unusable:

Screenshot from 2022-12-02 13-35-14

After my patch, it works fine with GNOME Shell's panel:

Screenshot from 2022-12-02 13-40-54

(This is also what it used to be under GNOME Wayland before <https://github.com/google/mozc/commit/310529c030a2cae697d72068b2ad76f9ec3f5325> enables Qt renderer under Wayland.)

This also works under X11 and looks better because it's the same style with GNOME itself:

photo_2022-12-02_14-08-37

AlynxZhou:
If there are any problems of this PR please comment, I am glad to fix them, thanks!

AlynxZhou:
Or if mozc really needs to use its own panel, at least revert <https://github.com/google/mozc/commit/310529c030a2cae697d72068b2ad76f9ec3f5325>, because the Qt renderer does not actually work with Wayland.

I also suggest to check some ENV (e.g. `MOZC_NO_RENDERER`) so user can choose between GNOME Shell's panel and mozc's own panel. Currently I can pass `--nouse_mozc_renderer` to `ibus-engine-mozc`, but this requires to edit the engine file of ibus, which will be replaced on system upgrade. ENV is easier to custom by users.

hiroyuki-komatsu:
Hi AlynxZhou,

Thank you for your PR. As you noticed, we do not accept the PR to the unix/ibus/ directory yet.

For your information, after we stop supporting the IBus client build with GYP, we are going to accept PRs.
* https://github.com/google/mozc/issues/567


Talking about this problem, the root cause from this project's viewpoint is that IBusEngine.cursor_area does not contain actual cursor information in Wayland. So we hope IBus will support the cursor information in Wayland too.
* https://ibus.github.io/docs/ibus-1.5/IBusEngine.html#IBusEngine-struct

That said, introducing an environment variable or some similar ways make sense as a workaround.
We are considering a good way to do it.

Thank you for your suggestion.

hiroyuki-komatsu:
We introduced the MOZC_IBUS_CANDIDATE_WINDOW environment variable as you suggested.
If you set it to `ibus`, you can use the Ibus default candidate window or behavior.

https://github.com/google/mozc/commit/4a77c6a7a8d569a93310e5b0277a5763b3b987e3


I also understand the issue of setting the window position in Wayland.
Thank you for your explanation.

I will try to make time to investigate how ibus and other IMFs (e.g. fcitx) set their window position in Wayland.
If you know some references, would you share them with us?

Best regards,

PR #643, opened 2022-12-09
(mozc.el) Fix warnings due to cl library deprecation

odanoburu:
Use `cl-lib` instead, which does not pollute the Emacs namespace. `cl-lib` is available since at least since Emacs 25.1 from six years ago (we are at Emacs 28.2, release 29.1 is due soon). The oldest Ubuntu release still supported (18.04) has Emacs 25.1, so I think it's safe to merge this, but if you want to support Emacs < 25 then this PR should be closed.

### DIFF
--- a/src/unix/emacs/mozc.el
+++ b/src/unix/emacs/mozc.el
@@ -120,7 +120,7 @@
 
 
 (eval-when-compile
-  (require 'cl))
+  (require 'cl-lib))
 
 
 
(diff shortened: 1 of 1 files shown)


hiroyuki-komatsu:
Hi odanoburu,

We have merged your commit as https://github.com/google/mozc/commit/cebb883f5d7d8d6dec05c6e2ed4096d0e8ff8c9f
Thank you,

Issue #701, opened 2023-03-23
Typing issue: 圃場


nus:
**Category of the typing issue**
Choose one of them (delete rest of them)
1. Out-of-vocabulary (e.g. "夕日" is not in the candidate list at all).

**Issues**
Write issues to the following table. (It's in the markdown format)

| input [e.g.ゆうひ] | expected [e.g. 夕日] | actual [e.g. ユウヒ] |
| ------------ | ------------ | ------------|
| ほじょう | 圃場 | ほ場 |

"ほ場" だけでなく漢字のみでの表記も加えていただきたいです。

**Version or commit-id**
Mozc-2.28.5029.101+24.11.oss

**Additional context**

hiroyuki-komatsu:
PR#703 added this entry.
Thank you!

PR #703, opened 2023-03-26
Add 圃場(ほじょう)

nus:
* #701

#typingquality


hiroyuki-komatsu:
Hi nus, thank you for you PR!

I just noticed that `evalution.tsv` is not a file we can accept PRs yet. It should be acceptable though.

Would you change this PR to have only `aux_dictionary.tsv` at this morment?
Or would you wait for a while until I change `evaluation.tsv` acceptable?

Thank you,

nus:
@hiroyuki-komatsu 
Thank you for the review!

Sure. I have changed this PR only with `aux_dictionary.tsv`.
Would you re-review?

nus:
Thanks!

PR #744, opened 2023-05-14
(mozc.el) Fix overlay not displaying in emacs

k0aaa:
The candidate overlay does not display in emacs even if `mozc-candidate-style` is set to `overlay`, and falls back to `echo-area`. This happens due to the non-existence in Elisp of the shorthands `first` and `third` for list manipulation at the relevant places.

Switching to using `nth` resolves the problem.


hiroyuki-komatsu:
Thank you for your PR.
We merged it to the repository.

PR #785, opened 2023-08-03
Remove needless quotes from case key-list

syohex:
**Description**

Remove needless quote from case key-list. This causes byte-compile warnings(as followings) since emacs 29.1. If the key is quoted, it matches against key and `quote`. For example `'eisu-toggle` matches `eisu-toggle` and `quote` symbol.

```
mozc.el:394:25: Warning: Case 'eisu-toggle will match ‘quote’.  If that’s
    intended, write (eisu-toggle quote) instead.  Otherwise, don’t quote
    ‘eisu-toggle’.
mozc.el:394:25: Warning: Case 'hiragana-katakana will match ‘quote’.  If
    that’s intended, write (hiragana-katakana quote) instead.  Otherwise,
    don’t quote ‘hiragana-katakana’.
...
```

**Issue IDs**

**Modified code locations**

- src/unix/emacs/mozc.el

**Confirmation of the acceptable code locations**

"Yes"

**Steps to test new behaviors (if any)**

- Install Emacs 29.1 or higher versions
- Byte-compile
- Check if the warnings are gone

**Additional context**

### DIFF
--- a/src/unix/emacs/mozc.el
+++ b/src/unix/emacs/mozc.el
@@ -395,37 +395,37 @@ Key code and symbols are renamed so that the helper process understands them."
                  (?\b 'backspace)
                  (?\s 'space)
                  (?\d 'backspace)
-                 ('eisu-toggle 'eisu)
-                 ('hiragana-katakana 'kana)
-                 ('next 'pagedown)
-                 ('prior 'pageup)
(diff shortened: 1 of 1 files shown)


hiroyuki-komatsu:
Hi syohex,

We are going to merge your PR at around our next update.
Thank you,

PR #787, opened 2023-08-05
Specify minimum emacs version

syohex:
## Description

Specify minimum emacs version because cl-lib has been bundled since Emacs 24.3(that was released in 2013 March).

And replace mozc-string-match-p with string-match-p. string-match-p was introduced at version 23. And `inhibit-changing-match-data` is obsoleted from Emacs 29.1.

```
In mozc-string-match-p:
mozc.el:1762:10: Warning: ‘inhibit-changing-match-data’ is an obsolete
    variable (as of 29.1);
```

## Modified code locations

- src/unix/emacs/mozc.el

## Confirmation of the acceptable code locations

"Yes"

### DIFF
--- a/src/unix/emacs/mozc.el
+++ b/src/unix/emacs/mozc.el
@@ -3,6 +3,7 @@
 ;; Copyright 2010-2021, Google Inc. All rights reserved.
 
 ;; Keywords: mule, multilingual, input method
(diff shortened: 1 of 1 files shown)


hiroyuki-komatsu:
Hi syohex,

We have merged your PR to the repository.
Thank you for your support.

PR #830, opened 2023-10-19
Zombie Process Prevention

phoepsilonix:
## Description
If you start and close the mozc_tool,
there are cases where the zombie process remains,
so prevent it.

## Modified code locations
src/base/process.cc

## Confirmation of the acceptable code locations
"Yes"

### DIFF
--- a/src/base/process.cc
+++ b/src/base/process.cc
@@ -213,6 +213,19 @@ bool Process::SpawnProcess(const std::string &path, const std::string &arg,
 #endif  // __linux__
   pid_t tmp_pid = 0;
 
+#ifdef __linux__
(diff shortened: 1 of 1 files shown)


yukawa:
Thank you for the pull request, but unfortunately `src/base/process.cc` isn't yet included in the list of files and directories that we are able to accept pull requests as of writing.
https://github.com/google/mozc/blob/af209061632ba7dcc64b63bd06eb988974375508/CONTRIBUTING.md?plain=1#L23-L38

Can you start a discussion thread first at https://github.com/google/mozc/discussions ?
To me, the right long term solution would be to stop relying on `posix_spawn`.

Issue #948, opened 2024-06-19
Use bazel to build Mozc for Windows


yukawa:
## Description
This is a placeholder issue to keep track of remaining tasks to migrate to Bazel in Windows build.

## Version or commit-id
07ce4568eb2ecf0deef60f83e90c0af696bc5048

yukawa:
Closing as we can at least build `Mozc64.msi`, which is basically usable.

Remaining tasks (e.g. migrating compile / link options such as `/LTCG`, `/DEPENDENTLOADFLAG`, `/CETCOMPAT`, updating the documentation) will be handled in separate issues/tasks.

PR #1057, opened 2024-09-05
Use `vswhere` to find Visual C++ in Windows

tats-u:
## Description
A clear and concise description of this pull request.

Visual Studio provides `vswhere` to make it easier to find the path of a file provided by Visual Studio.
We should use it to find `vcvarsall.bat` instaed of designating its path directly.

## Issue IDs
Issue and/or discussion IDs related to this pull request.

 - OS: Windows
 - Steps:
   1. `cd src`
   2. Run sub commands of `python ./build_mozc.py`

## Additional context

We won't need to make extra changes for newer Visual Studio in the future anymore.

### DIFF
--- a/src/build_tools/vs_util.py
+++ b/src/build_tools/vs_util.py
@@ -33,8 +33,23 @@ import json
 import pathlib
 import subprocess
 import sys
-from typing import Union
+from typing import Union, Optional
 
+def find_vswhere() -> Optional[pathlib.Path]:
+  path = pathlib.Path(r'C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe')
+  return path if path.exists() else None
+
+def get_vcvarsall_from_vswhere(vswhere_path: pathlib.Path) -> Optional[pathlib.Path]:
+  result = subprocess.run([str(vswhere_path), '-find', 'VC/Auxiliary/Build/vcvarsall.bat', '-utf8'], stdout=subprocess.PIPE)
+  first = result.stdout.splitlines()[0].decode()
+  if first == '':
+    return None
+  return pathlib.Path(first)
+
+def get_display_name_from_vswhere(vswhere_path: pathlib.Path) -> Optional[str]:
+  result = subprocess.run([str(vswhere_path), '-property', 'displayName', '-utf8'], stdout=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


yukawa:
Thank you for your pull request. While I agree that `vswhere` is a nice tool we can rely on, I’m leaning towards revisiting this after we complete the migration to bazel as being tracked in #948. Feel free to resend the pull request once #948 is completed and `build_mozc.py` is completely removed.

> We won't need to make extra changes for newer Visual Studio in the future anymore.

It’s highly unlikely that we can support any newer version of Visual Studio as long as we keep using GYP, which is no longer actively developed.

src/build_tools/vs_util.py
```diff
@@ -33,8 +33,23 @@
 import pathlib
 import subprocess
 import sys
-from typing import Union
+from typing import Union, Optional
 
+def find_vswhere() -> Optional[pathlib.Path]:
+  path = pathlib.Path(r'C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe')
```
yukawa, line 39: Let’s use `%ProgramFiles(x86)%`.

src/build_tools/vs_util.py
```diff
@@ -33,8 +33,23 @@
...
+
+def get_vcvarsall_from_vswhere(vswhere_path: pathlib.Path) -> Optional[pathlib.Path]:
+  result = subprocess.run([str(vswhere_path), '-find', 'VC/Auxiliary/Build/vcvarsall.bat', '-utf8'], stdout=subprocess.PIPE)
+  first = result.stdout.splitlines()[0].decode()
+  if first == '':
+    return None
+  return pathlib.Path(first)
+
+def get_display_name_from_vswhere(vswhere_path: pathlib.Path) -> Optional[str]:
+  result = subprocess.run([str(vswhere_path), '-property', 'displayName', '-utf8'], stdout=subprocess.PIPE)
```
yukawa, line 50: Unless it’s absolutely necessary, I prefer to set `shell=False` to avoid any unnecessary complexities coming from `cmd.exe`.

tats-u:
> build_mozc.py is completely removed.

`vs_uti.py` will be also unnecessary after that, I think.

Issue #853, opened 2023-12-19
Full-width Space Inserted Instead of Language Switch on Pressing Super+Space in Hiragana Input Mode in Fedora 39


earlydogs:
## Description
In Fedora 39, when the input mode is set to Hiragana and the language switch key (Super+Space) is pressed, it fails to switch languages and instead inserts a full-width space. This issue occurs during input in Terminal applications but does not happen in browser applications like Chrome. Initially thought to be a Fedora 39 bug, this issue does not occur with other input sources (such as Anthy), leading to this report.

## Steps to reproduce
1. Set up Fedora 39.
2. Install ibus-mozc with sudo dnf install ibus-mozc.
3. Restart the operating system.
4. Add Japanese (Mozc) to Input sources through the Settings menu.
5. Switch to Mozc by pressing Super + Space.
6. Change the Mozc setting to Hiragana in Input Mode.
7. In a terminal application, press Super + Space.

## Version or commit-id
Installed Packages
Name         : ibus-mozc
Version      : 2.29.5111.102
Release      : 3.fc39
Architecture : x86_64
Size         : 1.1 M
Source       : mozc-2.29.5111.102-3.fc39.src.rpm
Repository   : @System
From repo    : fedora
Summary      : The mozc engine for IBus input platform
URL          : https://github.com/google/mozc
License      : BSD-3-Clause AND Apache-2.0 AND Unicode-DFS-2015 AND NAIST-2003
Description  : Mozc is a Japanese Input Method Editor (IME) designed for
             : multi-platform such as Chromium OS, Windows, Mac and Linux.
             : 
             : This package contains the Input Method Engine for IBus.

Installed Packages
Name         : mozc
Version      : 2.29.5111.102
Release      : 3.fc39
Architecture : x86_64
Size         : 23 M
Source       : mozc-2.29.5111.102-3.fc39.src.rpm
Repository   : @System
From repo    : fedora
Summary      : A Japanese Input Method Editor (IME) designed for multi-platform
URL          : https://github.com/google/mozc
License      : BSD-3-Clause AND Apache-2.0 AND Unicode-DFS-2015 AND NAIST-2003
Description  : Mozc is a Japanese Input Method Editor (IME) designed for
             : multi-platform such as Chromium OS, Windows, Mac and Linux.

## Environment
 - OS: Fedora 39

PR #1059, opened 2024-09-13
Ignore Super and other modifier keys except for Alt

kzys:
## Description

Mozc doesn't know Super and other extra modifier keys except for Alt. For example, "Super + Space" (IBus' default for switching input methods) is recognized as "Space".
    
This change ignores key events that have these extra modifiers to let others handle the events.

## Issue IDs

#853

 - OS: Ubuntu 24.04 + Wayland
 - Steps:
   1. Switch to Japanese (Mozc) by pressing Super-Space
   2. Switch to English by pressing Super-Space again
   3. Confirm the switch (before this PR,  a full-width space was inserted instead)

## Additional context

### DIFF
--- a/src/unix/ibus/key_event_handler.cc
+++ b/src/unix/ibus/key_event_handler.cc
@@ -62,6 +62,18 @@ bool KeyEventHandler::GetKeyEvent(uint keyval, uint keycode, uint modifiers,
   DCHECK(key);
   key->Clear();
 
+  // Ignore key events with modifiers, except for the below;
+  // - Alt (Mod1) - Mozc uses Alt for shortcuts
+  // - NumLock (Mod2) - NumLock shouldn't impact shortcuts
+  // This is needed for handling shortcuts such as Super (Mod4) + Space,
+  // IBus's default for switching input methods.
+  // https://github.com/google/mozc/issues/853
+  constexpr uint kExtraModMask =
+      IBUS_MOD3_MASK | IBUS_MOD4_MASK | IBUS_MOD5_MASK;
+  if (modifiers & kExtraModMask) {
+    return false;
+  }
+
   if (!key_translator_->Translate(keyval, keycode, modifiers, preedit_method,
                                   layout_is_jp, key)) {
     LOG(ERROR) << "Translate failed";
--- a/src/unix/ibus/key_event_handler_test.cc
+++ b/src/unix/ibus/key_event_handler_test.cc
@@ -180,6 +180,13 @@ TEST_F(KeyEventHandlerTest, GetKeyEvent) {
     EXPECT_NO_MODIFIERS_PRESSED();
   }
 
+  { // Ignore Super (Mod4)
+    key.Clear();
+    EXPECT_FALSE(handler_->GetKeyEvent(IBUS_space, kDummyKeycode,
+                                       IBUS_MOD4_MASK, config::Config::ROMAN,
+                                       true, &key));
+  }
+
   // This test fails in current implementation.
   // TODO(hsumita): Enables it.
   /*


src/unix/ibus/key_event_handler.cc
```diff
@@ -38,6 +38,10 @@ namespace mozc {
 namespace ibus {
 
 namespace {
+
+// Mod1 is Alt, which is used by Mozc.
+const uint kExtraModMask = IBUS_MOD2_MASK | IBUS_MOD3_MASK | IBUS_MOD4_MASK | IBUS_MOD5_MASK;
```
hiroyuki-komatsu, line 43: Let's use constexper instead of const, and move this into the function.
Also please wrap the line within 80 chars.

```
  // Mod1 is Alt, which is used by Mozc.
  constexpr uint kExtraModMask =
      IBUS_MOD2_MASK | IBUS_MOD3_MASK | IBUS_MOD4_MASK | IBUS_MOD5_MASK;
  if (modifiers & kExtraModMask) {
    return false;
  }
```

phoepsilonix:
When using ibus-mozc with Japanese input mode enabled and set to full-width hiragana input mode, there is an issue where pressing Super+space inserts a full-width space instead of switching the input source. This PR resolves this issue. I tested it on Manjaro Linux and confirmed that it works well.

src/unix/ibus/key_event_handler.cc
```diff
@@ -62,6 +66,14 @@ bool KeyEventHandler::GetKeyEvent(uint keyval, uint keycode, uint modifiers,
   DCHECK(key);
   key->Clear();
 
+  // If the key is pressed with these extra modifiers that Mozc doesn't know,
+  // don't pass to the translator.
+  // This is needed to handle Super-Space, IBus's default for switching input methods.
```
hiroyuki-komatsu, line 71: Please wrap the line within 80 chars.

```
  // If the key is pressed with these extra modifiers that Mozc doesn't know,
  // don't pass to the translator. This is needed to handle Super-Space, IBus's
  // default for switching input methods.
  // https://github.com/google/mozc/issues/853
```

hiroyuki-komatsu:
Hi kzys,

Thank you for your PR with tests! Overall, it looks good to me. 
I left small comments regarding our coding style.

Thank you,

kzys:
Thanks! Addressed in the last revision.

phoepsilonix:
I'm encountering some issues. When I checked earlier, there were no problems, but now, upon rechecking, I've found the following issues. While switching input sources works fine, I'm facing two problems:

    The system doesn't accept input of full-width characters like hiragana.
    Pressing the full-width key doesn't switch the input mode.

Currently, I'm clearing the cache and rebuilding to recheck these issues.

phoepsilonix:
After testing in my local environment, I found that either of the following solutions resulted in the expected behavior:

1.  Completely remove the changes in src/unix/ibus/key_event_handler.cc and keep only the changes in src/unix/ibus/key_event_handler_test.cc.
2.  In the modified part of src/unix/ibus/key_event_handler.cc, set kExtraModMask to only IBUS_MOD4_MASK.

src/unix/ibus/key_event_handler.cc
```diff
@@ -38,6 +38,7 @@ namespace mozc {
 namespace ibus {
 
 namespace {
+
```
hiroyuki-komatsu, line 41: Please remove this line.

phoepsilonix:
I continued troubleshooting.
It might depend on keyboard layouts, but in my environment, adding IBUS_MOD2_MASK to kExtraModMask prevented the IME from being activated by the half-width/full-width key or the hiragana key.
Removing IBUS_MOD2_MASK from kExtraModMask and using only IBUS_MOD4_MASK, or adding either or both of IBUS_MOD3_MASK and IBUS_MOD5_MASK, allowed the IME to be activated.
Since this might be environment-dependent, I think it might be better to limit it to just IBUS_MOD4_MASK for the Super key to minimize potential impacts.

For your reference.

```
 * @IBUS_MOD1_MASK: Modifier 1 (Usually Alt_L (0x40),  Alt_R (0x6c),  Meta_L (0xcd)) activated.
 * @IBUS_MOD2_MASK: Modifier 2 (Usually Num_Lock (0x4d)) activated.
 * @IBUS_MOD3_MASK: Modifier 3 activated.
 * @IBUS_MOD4_MASK: Modifier 4 (Usually Super_L (0xce),  Hyper_L (0xcf)) activated.
 * @IBUS_MOD5_MASK: Modifier 5 (ISO_Level3_Shift (0x5c),  Mode_switch (0xcb)) activated.
```

src/unix/ibus/key_event_handler.cc
kzys: Will do.

kzys:
>  It might depend on keyboard layouts, but in my environment, adding IBUS_MOD2_MASK to kExtraModMask prevented the IME from being activated by the half-width/full-width key or the hiragana key.

Thanks. Interesting. Let me check. I don't have a Japanese keyboard nearby though.

I don't know how these Japanese-specific keys get Mod1. Is it coming from XKB somehow?

https://github.com/ibus/ibus/blob/fa10c116c2f0dad5026fa6665e9513cd5dcded75/client/wayland/ibuswaylandim.c#L690-L691

kzys:
@phoepsilonix Can you check `xmodmap` on your machine? I got the below.

```
% xmodmap 
xmodmap:  up to 4 keys per modifier, (keycodes in parentheses):

shift       Shift_L (0x32),  Shift_R (0x3e)
lock        Caps_Lock (0x42)
control     Control_L (0x25),  Control_R (0x69)
mod1        Alt_L (0x40),  Alt_R (0x6c),  Alt_L (0xcc),  Meta_L (0xcd)
mod2        Num_Lock (0x4d)
mod3        ISO_Level5_Shift (0xcb)
mod4        Super_L (0x85),  Super_R (0x86),  Super_L (0xce),  Hyper_L (0xcf)
mod5        ISO_Level3_Shift (0x5c)
```

phoepsilonix:
The differences are about as follows. Eisu_toglle probably refers to the alphanumeric toggle.

lock        Eisu_toggle (0x42)
lock        Caps_Lock (0x42)

```sh
xmodmap:  up to 4 keys per modifier, (keycodes in parentheses):

shift       Shift_L (0x32),  Shift_R (0x3e)
lock        Eisu_toggle (0x42)
control     Control_L (0x25),  Control_R (0x69)
mod1        Alt_L (0x40),  Alt_R (0x6c),  Alt_L (0xcc),  Meta_L (0xcd)
mod2        Num_Lock (0x4d)
mod3        ISO_Level5_Shift (0xcb)
mod4        Super_L (0x85),  Super_R (0x86),  Super_L (0xce),  Hyper_L (0xcf)
mod5        ISO_Level3_Shift (0x5c)
```

phoepsilonix:
I use NUMLOCK enabled at startup. I always keep NUMLOCK on because I have a numeric keypad. Some keyboards don't have a numeric keypad, and NUMLOCK might affect the main keyboard layout. I was wondering if this could be the cause of the issue. So I tested what happens when NUMLOCK is turned off.
This is a behavior test of ibus-mozc with IBUS_MOD2_MASK added to kExtraModMask.
1. When NUMLOCK is off, the IME becomes active and Japanese input becomes possible.
2. When NUMLOCK is on, it seems impossible to toggle the IME on and off.
In other words, the on/off state of NUMLOCK is affecting the behavior of the full-width/half-width key and the hiragana key. This issue does not occur with ibus-mozc without adding IBUS_MOD2_MASK to kExtraModMask.

kzys:
@phoepsilonix Thanks! That makes more sense. Updated the mask.

src/unix/ibus/key_event_handler.cc
```diff
@@ -62,6 +62,18 @@ bool KeyEventHandler::GetKeyEvent(uint keyval, uint keycode, uint modifiers,
...
   key->Clear();
 
+  // Ignore key events with modifiers, except for the below;
+  // - Alt (Mod1) - Mozc uses Alt for shortcuts
+  // - NumLock (Mod2) - NumLock shouldn't impact shortcuts
+  // This is needed for handling shortcuts such as Super (Mod4) + Space,
+  // IBus's default for switching input methods.
+  // https://github.com/google/mozc/issues/853
+  constexpr uint kExtraModMask =
+    IBUS_MOD3_MASK | IBUS_MOD4_MASK | IBUS_MOD5_MASK;
```
yukawa, line 72: The line 72 should have 4 spaces for the indentation.
```
  constexpr uint kExtraModMask =
      IBUS_MOD3_MASK | IBUS_MOD4_MASK | IBUS_MOD5_MASK;
```

kzys: I missed that :(

Fixed in the last revision.

hiroyuki-komatsu:
We have merged your PR.

PR #1063, opened 2024-09-22
Modification of the postal code dictionary script.

phoepsilonix:
## Description
Modification of the postal code dictionary script.

    Remove redundant entries.
    
    * Examples
      + 琴平町の次に４２７番地以降がくる場合（川西） : 766-0001 香川県仲多度郡琴平町
      + 琴平町の次に１～４２６番地がくる場合（川東） : 766-0002 香川県仲多度郡琴平町

    #typingquality

Please try to convert the following postal code.
766-0001
766-0002


src/dictionary/gen_zip_code_seed.py
```diff
@@ -129,8 +129,10 @@ def ParseTownName(level3):
   # 409-0142 小菅村の次に1~663番地がくる場合 (山梨県北都留郡小菅村)
   # 409-0211 小菅村の次に664番地以降がくる場合 (山梨県北都留郡小菅村)
   # 901-3601 渡名喜村一円 (沖縄県島尻郡渡名喜村)
+  # 766-0001 琴平町の次に４２７番地以降がくる場合（川西）
```
hiroyuki-komatsu, line 132: Please append " (香川県仲多度郡琴平町) " to the line for the consistency.

```
# 766-0001 琴平町の次に４２７番地以降がくる場合（川西） (香川県仲多度郡琴平町) 
```

It's also fine to reformat all lines, like:

```
# 中津市の次に番地がくる場合 : 871-0099 大分県中津市
...
# 琴平町の次に４２７番地以降がくる場合（川西） : 766-001 香川県仲多度郡琴平町
```

hiroyuki-komatsu:
Thank you for your PR!

It looks good to me. Please check my comment for format.

src/dictionary/gen_zip_code_seed.py
phoepsilonix: I fixed it as above. What do you think?

phoepsilonix: This is a separate matter, but since it's about the same script, let me ask a question.

The purpose of using 'from dictionary import zip_code_util' is to make it easier to find the location of zip_code_util when called from within the Bazel sandbox, right?
How about the following change?
```diff
diff --git a/src/dictionary/BUILD.bazel b/src/dictionary/BUILD.bazel
index 08ef57d..b91103c 100644
--- a/src/dictionary/BUILD.bazel
+++ b/src/dictionary/BUILD.bazel
@@ -729,6 +729,7 @@ mozc_py_binary(
         "gen_zip_code_seed.py",
         "zip_code_util.py",
     ],
+    imports = ["."],
     visibility = ["//train/zip_code:__subpackages__"],
 )
 
diff --git a/src/dictionary/gen_zip_code_seed.py b/src/dictionary/gen_zip_code_seed.py
index 69bdffa..597b165 100755
--- a/src/dictionary/gen_zip_code_seed.py
+++ b/src/dictionary/gen_zip_code_seed.py
@@ -57,7 +57,7 @@ import re
 import sys
 import unicodedata
 
-from dictionary import zip_code_util
+import zip_code_util
 
 
 ZIP_CODE_LABEL = 'ZIP_CODE'
```
By instructing imports in mozc_py_binary, I think it will be able to find items in the same folder.
And I think when calling the gen_zip_code_seed.py script from the shell, it can be executed without adding the path of Mozc's source root directory(mozc/src/ or $PWD) to PYTHONPATH. 
What do you think?

hiroyuki-komatsu: > Would it be better to create a separate PR rather than adding commits to this PR?

Yes, we prefer a minimal and single matter per PR.
Thank you,

phoepsilonix: I understand.

hiroyuki-komatsu: GitHub Discussions is also preferred as well as PR.

For your question, we would not add `imports = ["."]` to the rules in general for the consistency to save the maintenance efforts.

To avoid specifying `PYTHONPATH`, you can also use the file built by Bazel.

```
bazel build dictionary:gen_zip_code_seed --config oss_linux
./bazel-bin/dictionary/gen_zip_code_seed --help
```

phoepsilonix: I see.

Let's get back to the main topic.

hiroyuki-komatsu:
We have merged your PR.

PR #1064, opened 2024-09-23
Simplify the Python script by adjusting the Bazel configuration file.

phoepsilonix:
## Description
The purpose of using 'from dictionary import zip_code_util' is to make it easier to find the location of zip_code_util when called from within the Bazel sandbox, right?

By instructing imports in mozc_py_binary, I think it will be able to find items in the same folder.
And I think when calling the gen_zip_code_seed.py script from the shell, it can be executed without adding the path of Mozc's source root directory(mozc/src/ or $PWD) to PYTHONPATH. 

The zip_code.tsv file is created without any issues using the bazel build command.
```sh
bazel build --config oss_linux --config release_build dictionary:zip_code_data --verbose_failures --sandbox_debug
```
There is no need to specify PYTHONPATH when using it in the shell.
```sh
python ../dictionary/gen_zip_code_seed.py  --zip_code=KEN_ALL.CSV --jigyosyo=JIGYOSYO.CSV > zip_code_seed.tsv
```

### DIFF
--- a/src/dictionary/BUILD.bazel
+++ b/src/dictionary/BUILD.bazel
@@ -729,6 +729,7 @@ mozc_py_binary(
         "gen_zip_code_seed.py",
         "zip_code_util.py",
     ],
+    imports = ["."],
     visibility = ["//train/zip_code:__subpackages__"],
 )
 
--- a/src/dictionary/gen_zip_code_seed.py
+++ b/src/dictionary/gen_zip_code_seed.py
@@ -35,7 +35,6 @@ r"""Zip code dictionary generator.
  Output lines will be printed as utf-8.
 
  usage:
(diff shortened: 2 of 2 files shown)


hiroyuki-komatsu:
Hi again, thank you for your PR!

We would not add `imports = ["."]` to the rules in general for the consistency to save the maintenance efforts.

To avoid specifying `PYTHONPATH`, you can also use the file built by Bazel.

```
bazel build dictionary:gen_zip_code_seed --config oss_linux
./bazel-bin/dictionary/gen_zip_code_seed --help
```

This way is available for all `*_binary` Bazel rules (e.g. `mozc_py_binary`).

phoepsilonix:
I understand.

Issue #1076, opened 2024-10-12
[Windows] SearchHost.exe may freeze caused by FileUtil::CreateDirectory


hiroyuki-komatsu:
There are some reports that SearchHost.exe freezes with Mozc built with `BRANDING=GoogleJapaneseInput`.

Through bisect investigations, I figured out 68e99df9805875b1a5b9906652aa967678ac59bf is the change of this issue. It modified FileUtil::CreateDirectory.

We should update FileUtil::CreateDirectory again to address the problem.

As a tricky point, this issue is reported only on `BRANDING=GoogleJapaneseInput` and 'channel_dev=0`.

References:
* https://support.google.com/gboard/thread/300465502?hl=ja&msgid=301486087
* https://answers.microsoft.com/ja-jp/windows/forum/all/windowssearchhost%E3%81%8C%E4%B8%AD%E6%96%AD/03d7f11c-f73d-4b0f-b161-e4b90dd371c4

yukawa:
While I still cannot reproduce the issue locally, based on the information I got my current hypothesis is that `broadFileSystemAccess` capability given to `SearchHost.exe` is playing an interesting role.

When Mozc TIP DLL calls `CreateFileW` API to open `config1.db`, the process itself does not have sufficient  permission to open it. Then `broadFileSystemAccess` capability will take place and `Windows.Storage.OneCore.dll` will attempts brokered file access after initializing the thread with `RoInitialize()` when not yet initialized. If this happens in a certain situation, TSF runtime gets confused and may start re-initializing Mozc TIP.

  1. TSF calls back into Mozc TIP
  2. Mozc TIP calls `CreateFileW()`
  3. The system invokes `RoInitialize()` before returning from `CreateFileW()`
  4. TSF calls back into Mozc TIP before returning from `RoInitialize()`
  5. Now Mozc TIP is handling a reentrant callback.

The on-going issue looks to be a deadlock due to non-reentrant lock.

  1. TSF calls back into Mozc TIP
  2. Mozc TIP acquires a non-reentrant lock.
  3. Mozc TIP calls `CreateFileW()`
  4. The system invokes `RoInitialize()` before returning from `CreateFileW()`
  5. TSF calls back into Mozc TIP before returning from `RoInitialize()`
  6. Now Mozc TIP is handling a reentrant callback.
  7. Mozc TIP tries to acquire the same non-reentrant lock, which causes a dead lock.

yukawa:
[...]

Issue #856, opened 2023-12-22
Applications that do not explicitly call `CoInitialize` may crash when loading or enabling Mozc in Windows


yukawa:
## Description
When an application that does not explicitly call `CoInitialize` try to load Mozc's dll (or the user tries to enable Mozc), the application may crash due to system's unexpectedly unloading Mozc's dll.

## Steps to reproduce
1. Install Mozc
2. Build and launch the following app.
3. Select Mozc

```
#include <windows.h>

int WINAPI wWinMain(HINSTANCE instance_handle, HINSTANCE prev_instance_handle,
                    PWSTR command_line, int command_show) {
  const wchar_t window_class_name[] = L"Test Window Class";
  {
    const WNDCLASS wc = {
        .lpfnWndProc = [](HWND window_handle, UINT message,
                          auto... params) -> LRESULT {
          if (message == WM_DESTROY) {
            ::PostQuitMessage(0);
            return 0;
          }
          return ::DefWindowProc(window_handle, message, params...);
        },
        .hInstance = instance_handle,
        .hbrBackground = (HBRUSH)(COLOR_WINDOW + 1),
        .lpszClassName = window_class_name,
    };
    ::RegisterClass(&wc);
  }

  const HWND window_handle = ::CreateWindowEx(
      0, window_class_name, L"Window Title", WS_OVERLAPPEDWINDOW, CW_USEDEFAULT,
      CW_USEDEFAULT, CW_USEDEFAULT, CW_USEDEFAULT, nullptr, nullptr,
      instance_handle, nullptr);

  if (window_handle == nullptr) {
    return 0;
  }

  ::ShowWindow(window_handle, command_show);

  while (true) {
    MSG message = {};
    if (::GetMessage(&message, nullptr, 0, 0) <= 0) {
      break;
    }
    ::TranslateMessage(&message);
    ::DispatchMessage(&message);
  }

  return 0;
}
```

## Expected behavior
The test app does not crash at the step 3.

## Actual behavior
The test app crashes at the step 3.

## Version or commit-id
68fea8a05c189a7ff166b36b07eac434569f4cd2

2.29.5288.100 and later versions are considered to be affected.

## Environment
 - OS: Windows 11 22H2

### What's actually happening
1. When an app creates a Win32 window without initializing COM, CUAS (Cicero Unaware Application Support) automatically calls `CoInitialize` with also installing `CoRegisterInitializeSpy`, because TSF is built on top of COM.
[...]

PR #1080, opened 2024-10-14
Make `config1.db` readable from `AppContainer` apps

yukawa:
## Description
This attempts to avoid the deadlock issue discussed in #1076.

From what I can observe on Windows 11 23H2 (22631.4317), it looks to be dangerous for us to call File I/O Win32 APIs such as `CreateFile` from AppContainer processes unless we are confident that the target file/dir is accessible to the AppContainer process. When such an access is not allowed at the ACL level, the debugger shows that `Windows.Storage.OneCore.dll` tries to forward it to a broker process (e.g. via `Windows.Storage.OneCore.dll!BrokeredCreateFile2`). The issue is that Windows.Storage.OneCore.dll is a WinRT component, which means that just calling `CreateFile()` results in dynamically invoking `RoInitialize()` internally, which can confuse TSF runtime as we have already seen in #856.

To summarize, the safest option is to ensure that the target file/dir is always accessible to AppContainer so that `Windows.Storage.OneCore.dll` will not be involved if the file/dir is opened with `CreateFile()` from `MozcTip{32,64}.dll`.

If this commit is not sufficient to address #1076, we then need to take care of other File I/O APIs such as `CreateDirectoryW()` and `GetFileAttributes()`.

## Issue IDs
#1076.

A clear and concise description about how to verify new behaviors.
 - OS: Windows 11 23H2
 - Steps:
   1. Install a previous version of Mozc.
   2. Open the config dialog then update the config. Then save it.
   3. `cacls %LOCALAPPDATA%Low\Mozc\config1.db`
   4. Create and install `Mozc64.msi` with this commit to update Mozc.
   5. `cacls %LOCALAPPDATA%Low\Mozc\config1.db`

* At the step 3, there is no entry about `ALL APPLICATION PACKAGES`.
* At the step 5, there should be an entry about `ALL APPLICATION PACKAGES` as follows.
```
APPLICATION PACKAGE AUTHORITY\ALL APPLICATION PACKAGES:(special access:)
                                                       READ_CONTROL
                                                       SYNCHRONIZE
                                                       FILE_GENERIC_READ
                                                       FILE_READ_DATA
                                                       FILE_READ_EA
                                                       FILE_READ_ATTRIBUTES
```
Not yet confirmed to be a valid fix.

### DIFF
--- a/src/base/config_file_stream.cc
+++ b/src/base/config_file_stream.cc
@@ -38,6 +38,10 @@
 #include <string>
 #include <utility>
 
+#ifdef _WIN32
+#include <windows.h>
+#endif  // _WIN32
+
 #include "absl/container/flat_hash_map.h"
 #include "absl/log/check.h"
 #include "absl/log/log.h"
@@ -52,7 +56,7 @@
 #include "base/system_util.h"
 
 #ifdef _WIN32
-#include <windows.h>
+#include "base/win32/win_sandbox.h"
 #endif  // _WIN32
 
 namespace mozc {
@@ -218,4 +222,22 @@ std::string ConfigFileStream::GetFileName(const absl::string_view filename) {
 void ConfigFileStream::ClearOnMemoryFiles() {
   Singleton<OnMemoryFileMap>::get()->clear();
 }
+
+#ifdef _WIN32
+// Check the file permission of "config1.db" if exists to ensure that
+// "ALL APPLICATION PACKAGES" have read access to it.
+void ConfigFileStream::FixupFilePermission(absl::string_view filename) {
+    const std::string path = ConfigFileStream::GetFileName(filename);
+    if (path.empty()) {
+      return;
+    }
+    const absl::Status status = FileUtil::FileExists(path);
+    if (status.ok()) {
+      WinSandbox::EnsureAllApplicationPackagesPermisssion(
+          win32::Utf8ToWide(path),
+          WinSandbox::AppContainerVisibilityType::kConfigFile);
+    }
+}
+#endif  // _WIN32
+
 }  // namespace mozc
--- a/src/base/config_file_stream.h
+++ b/src/base/config_file_stream.h
@@ -107,6 +107,10 @@ class ConfigFileStream {
   // Clear all memory:// files.  This is a utility method for testing.
   static void ClearOnMemoryFiles();
 
+#ifdef _WIN32
+  static void FixupFilePermission(absl::string_view filename);
+#endif  // _WIN32
+
  private:
   // This function is deprecated. Use OpenReadText or OpenReadBinary instead.
   // TODO(yukawa): Move this function to anonymous namespace in
--- a/src/base/win32/win_sandbox.cc
+++ b/src/base/win32/win_sandbox.cc
@@ -1015,6 +1015,23 @@ bool SetTokenIntegrityLevel(HANDLE token,
   return (result != FALSE);
 }
 
+constexpr ACCESS_MASK GetAccessMask(
+    WinSandbox::AppContainerVisibilityType type) {
+  const ACCESS_MASK kBaseType =
+      FILE_READ_DATA | FILE_READ_EA | READ_CONTROL | SYNCHRONIZE;
+
+  switch (type) {
+    case WinSandbox::AppContainerVisibilityType::kProgramFiles:
+      // As of Windows 10 Anniversary Update, following access masks (==0x1200a9)
+      // are specified to files under Program Files by default.
+      return FILE_EXECUTE | kBaseType;
+    case WinSandbox::AppContainerVisibilityType::kConfigFile:
+      return FILE_GENERIC_READ | FILE_READ_ATTRIBUTES | kBaseType;
+    default:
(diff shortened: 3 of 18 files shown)


src/base/config_file_stream.cc
```diff
@@ -218,4 +222,20 @@ std::string ConfigFileStream::GetFileName(const absl::string_view filename) {
 void ConfigFileStream::ClearOnMemoryFiles() {
   Singleton<OnMemoryFileMap>::get()->clear();
 }
+
+#ifdef _WIN32
+void ConfigFileStream::FixupFilePermission(absl::string_view filename) {
+    const auto path = ConfigFileStream::GetFileName(filename);
```
hiroyuki-komatsu, line 228: Let's use `std::string` instead of `auto`.

src/base/config_file_stream.h
```diff
@@ -107,6 +107,10 @@ class ConfigFileStream {
   // Clear all memory:// files.  This is a utility method for testing.
   static void ClearOnMemoryFiles();
 
+#ifdef _WIN32
+  static void FixupFilePermission(absl::string_view filename);
```
hiroyuki-komatsu, line 111: Please add a comment what this function does.

src/base/win32/win_sandbox.cc
```diff
@@ -1243,10 +1260,7 @@ bool WinSandbox::EnsureAllApplicationPackagesPermisssion(
     return false;
   }
 
-  // As of Windows 10 Anniversary Update, following access masks (==0x1200a9)
-  // are specified to files under Program Files by default.
-  const ACCESS_MASK kDesiredMask =
-      FILE_READ_DATA | FILE_READ_EA | FILE_EXECUTE | READ_CONTROL | SYNCHRONIZE;
+  const ACCESS_MASK kDesiredMask = GetAccessMask(type);
```
hiroyuki-komatsu, line 1263: How about renaming `kDesiredMask` to `desired_mask`?

src/win32/custom_action/custom_action.cc
```diff
@@ -328,6 +333,18 @@ UINT __stdcall EnableTipProfile(MSIHANDLE msi_handle) {
   return ERROR_SUCCESS;
 }
 
+UINT __stdcall FixupConfigFilePermission(MSIHANDLE msi_handle) {
```
hiroyuki-komatsu, line 336: Does this always return `ERROR_SUCCESS`? How about mentioning it to clarify the behavior?

hiroyuki-komatsu (commented):
Hi yukawa,
Thank you for your PR.

We appreciate your fix against #1076.

src/win32/tip/BUILD.bazel
```diff
@@ -437,6 +437,7 @@ mozc_cc_library(
     target_compatible_with = ["@platforms//os:windows"],
     deps = [
         "//protocol:commands_cc_proto",
+        "//client:client_interface",
```
hiroyuki-komatsu, line 440: Let's sort them.

```
"//client:client_interface",
"//protocol:commands_cc_proto",
"//win32/base:config_snapshot",
```

src/win32/tip/tip_private_context.h
```diff
@@ -35,6 +35,7 @@
 #include <memory>
 
 #include "protocol/commands.pb.h"
+#include "client/client_interface.h"
```
hiroyuki-komatsu, line 38: Let's sort them

```
#include "client/client_interface.h"
#include "protocol/commands.pb.h"
#include "win32/base/config_snapshot.h"
```

src/base/config_file_stream.cc
yukawa: Sorry, forgot to update it. Done in 353599a0a1015887c400ad3c0fb0198d978729c6.

src/base/config_file_stream.cc
```diff
@@ -52,7 +56,7 @@
 #include "base/system_util.h"
 
 #ifdef _WIN32
-#include <windows.h>
+#include "win32/win_sandbox.h"
```
hiroyuki-komatsu, line 59: `base/win32/win_sandbox.h` ?

yukawa: Ah, yes. Fixed.

src/win32/tip/tip_private_context.h
yukawa: Done.

src/base/win32/win_sandbox.cc
yukawa: Done.

src/win32/tip/BUILD.bazel
yukawa: Done.

src/base/config_file_stream.h
yukawa: Done.

src/win32/custom_action/custom_action.cc
yukawa: Done.

hiroyuki-komatsu:
We have merged your PR.

Issue #1086, opened 2024-10-17
Install Mozc into `%ProgramFiles%\Mozc` rather than `%ProgramFiles(x86)%\Mozc`


yukawa:
## Description
Historically we have been installing Mozc into `%ProgramFiles(x86)%\Mozc`  (e.g. `C:\Program Files (x86)\Mozc`) rather than `%ProgramFiles%\Mozc` even on 64-bint environment as that is how Google Japanese Input has been doing since 2009.

This is going to be much more confusing when we start supporting Windows Arm64 environment, where we may see the following 3 "Program Files" directories.

 * `C:\Program Files`: 64-bit binaries for both arm64 and x86_64
 * `C:\Program Files (Arm)`: arm 32-bit binaries
 * `C:\Program Files (x86)`: x86 32-bit binaries

Let's see if we can migrate from `%ProgramFiles(x86)%\Mozc` to `%ProgramFiles%\Mozc` without disturbing existing (Google Japanese Input) users.

## Steps to reproduce
1. Install some older version of Mozc, which is to be installed into `%ProgramFiles(x86)%\Mozc`.
2. Launch some app and select Mozc.
3. Build a new version of Mozc, which is to be installed into `%ProgramFiles%\Mozc`.
4. `msiexec.exe /i Mozc64.msi /qn REBOOT=ReallySuppress`

## Screenshots
Mozc continues working on the app launched at the step 2.

## Version or commit-id
8c8a4461febab02477b688e687765b67b6fd8927

## Environment
 - OS: Windows 11 23H2

## Additional context
 * https://github.com/google/mozc/issues/732
 * https://github.com/google/mozc/discussions/893

PR #1087, opened 2024-10-17
Retrieve Mozc's installation directory from registry

yukawa:
## Description
Currently the directory `%ProgramFiles(x86)%\Mozc` is effectively hard coded in `mozc_tip{32,64}.dll`. This is a big challenge when we want to switch Mozc's installation directory from `%ProgramFiles(x86)%\Mozc` to `%ProgramFiles%\Mozc`.

So as not to bother users during the migration, with this commit `SystemUtil::GetServerDirectory()` first attempts to retrieve Mozc's installation directory from registry by checking the COM entry for `mozc_tip64.dll`.

The next step is to deploy `mozc_tip{32,64}.dll` users with this commit.

## Issue IDs

* https://github.com/google/mozc/issues/1086

### DIFF
--- a/src/base/system_util.cc
+++ b/src/base/system_util.cc
@@ -409,11 +409,48 @@ class ProgramFilesX86Cache {
   HRESULT result_;
   std::string path_;
 };
+
+constexpr wchar_t kMozcTipClsid[] =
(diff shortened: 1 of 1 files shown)


hiroyuki-komatsu:
We have merged your PR.

PR #1093, opened 2024-10-23
Shutdown `mozc_{server,renderer}.exe` on upgrading

### LINKED ISSUE #1092: Consider shutting down `mozc_server.exe` and `mozc_renderer.exe` while updating Mozc
## Description
Currently `mozc_server.exe` and `mozc_renderer.exe` remain running when updating Mozc from an older version to a new version on Windows, just because our `ShutdownServer` custom action is configured to not run for upgrading scenarios.

https://github.com/google/mozc/blob/20b02d2b06c5dfab7aa55c34ba77edef01e81afd/src/win32/installer/installer_oss_64bit.wxs#L147-L153

https://github.com/google/mozc/blob/20b02d2b06c5dfab7aa55c34ba77edef01e81afd/src/win32/installer/installer_64bit.wxs#L171-L177

From the viewpoint of stability, we may most likely want to shut down them so that Windows Installer can replace existing files without having to rename them.

## Steps to reproduce
1. Install an older version of Mozc
2. Open some application (e.g. notepad) then select Mozc and type something with Mozc enabled.
3. Make sure that `mozc_server.exe` and `mozc_renderer.exe` are running.
4. Double click a newer version of `Mozc64.msi`

## Expected behavior
You are not asked to choose whether `mozc_server.exe` and `mozc_renderer.exe` should be restarted to complete the installation.

## Actual behavior
You are asked to choose whether `mozc_server.exe` and `mozc_renderer.exe` should be restarted to complete the installation.
Image

## Version or commit-id
20b02d2b06c5dfab7aa55c34ba77edef01e81afd

## Environment
 - OS: Windows 11 23H2

yukawa:
## Description
It would make more sense to shutdown `mozc_server.exe` and `mozc_renderer.exe` not only when uninstalling Mozc but also when upgrading Mozc.

With this commit

 * `ShutdownServer` custom actions will always run
 * Any failure on 'ShutdownServer' will not block installation/upgrading as it is an optional behavior and Windows Installer can proceed even when these processes continue running.

Closes #1092.

## Issue IDs

* https://github.com/google/mozc/issues/1092

 - OS: Windows 11 23H2
 - Steps:
   1. Install an older version of Mozc
   2. Open some application (e.g. notepad) then select Mozc and type something with Mozc enabled.
   3. Make sure that `mozc_server.exe` and `mozc_renderer.exe` are running.
   4. Double click a newer version of `Mozc64.msi`
   5. Make sure you are not asked to choose whether `mozc_server.exe` and `mozc_renderer.exe` should be restarted to complete the installation.

### DIFF
--- a/src/win32/custom_action/custom_action.cc
+++ b/src/win32/custom_action/custom_action.cc
@@ -244,21 +244,19 @@ UINT __stdcall ShutdownServer(MSIHANDLE msi_handle) {
   DEBUG_BREAK_FOR_DEBUGGER();
   std::unique_ptr<mozc::client::ClientInterface> server_client(
       mozc::client::ClientFactory::NewClient());
-  bool server_result = true;
   if (server_client->PingServer()) {
-    server_result = server_client->Shutdown();
+    if (!server_client->Shutdown()) {
+      // This is not fatal as Windows Installer can replace executables even
+      // when they still are running. Just log error then go ahead.
+      LOG_ERROR_FOR_OMAHA();
+    }
   }
   mozc::renderer::RendererClient renderer_client;
-  const bool renderer_result = renderer_client.Shutdown(true);
-  if (!server_result) {
+  if (!renderer_client.Shutdown(true)) {
+    // This is not fatal as Windows Installer can replace executables even when
+    // they are still running. Just log error then go ahead.
     LOG_ERROR_FOR_OMAHA();
-    return ERROR_INSTALL_FAILURE;
   }
-  if (!renderer_result) {
-    LOG_ERROR_FOR_OMAHA();
-    return ERROR_INSTALL_FAILURE;
-  }
-
   return ERROR_SUCCESS;
 }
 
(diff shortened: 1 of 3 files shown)


src/win32/custom_action/custom_action.cc
```diff
@@ -244,21 +244,15 @@ UINT __stdcall ShutdownServer(MSIHANDLE msi_handle) {
...
+  if (!renderer_client.Shutdown(true)) {
     LOG_ERROR_FOR_OMAHA();
-    return ERROR_INSTALL_FAILURE;
   }
-  if (!renderer_result) {
-    LOG_ERROR_FOR_OMAHA();
-    return ERROR_INSTALL_FAILURE;
-  }
-
   return ERROR_SUCCESS;
```
hiroyuki-komatsu, line 256: Please add a comment about why we can return SUCCESS here.

yukawa: Done.

PR #1095, opened 2024-10-26
Enabling Bazel to recognize packages that do not output cflags through pkg-config.

phoepsilonix:
Enabling Bazel to recognize packages that do not output cflags through pkg-config.

If a package's cflags only contain '-I/usr/include', there will be no output
from 'pkg-config --cflags-only-I'. Therefore, it does not recognize it as a
package. Additionally, the referenced headers (hdrs) become absolute paths
(`"/**"`), which causes Bazel to generate an error.
The glob pattern does not allow absolute paths(`"/**"`) or empty strings("").

This modification allows pkg-config to be re-executed with --keep-system-cflags
if result.stdout.strip() is empty ("").
Typically, -I/usr/include is returned, enabling bazel to recognize packages
as valid even when pkg-config does not output cflags with standard options.

## Description
Currently, when attempting to build fcitx-mozc with Bazel, the output of pkg-config --cflags-only-I is empty. As a result, the hdrs reference becomes an absolute path (/**), causing Bazel to stop due to not finding `@fcitx`.
If we can avoid this error, the build will be possible. Other packages may not have issues at present, but if cflags only includes -I/usr/include, there is a high likelihood of encountering the same type of error.

The bazel command returns an error with the following hdrs:
```
Error in glob: pattern cannot be absolute
```
```
    hdrs = glob([
        "/**"
    ]),
```
```
Error in glob: pattern cannot be empty
```
```
    hdrs = glob([
        ""
    ]),
```

This modification allows pkg-config to be re-executed with --keep-system-cflags if result.stdout.strip() is empty ("").
Typically, -I/usr/include is returned, enabling Bazel to recognize packages that do not output cflags through pkg-config.

### DIFF
--- a/src/bazel/pkg_config_repository.bzl
+++ b/src/bazel/pkg_config_repository.bzl
@@ -78,21 +78,14 @@ EXPORTS_FILES_TEMPLATE = """
 exports_files(glob(["libexec/*"]))
 """
 
-def _exec_pkg_config(repo_ctx, flag):
+def _exec_pkg_config(repo_ctx, flags):
     binary = repo_ctx.which("pkg-config")
     if not binary:
         # Using print is not recommended, but this will be a clue to debug build errors in
         # the case of pkg-config is not found.
         print("pkg-config is not found")  # buildifier: disable=print
         return []
-    result = repo_ctx.execute([binary, flag] + repo_ctx.attr.packages)
-
-    # If result.stdout.strip() is empty, pkg-config will be re-executed with
-    # the --keep-system-cflags option added. Typically, -I/usr/include is
-    # returned, enabling bazel to recognize packages as valid even when
-    # pkg-config does not output cflags with standard options.
-    if result.stdout.strip() == "":
-        result = repo_ctx.execute([binary, flag] + ["--keep-system-cflags"] + repo_ctx.attr.packages)
+    result = repo_ctx.execute([binary] + flags + repo_ctx.attr.packages)
     items = result.stdout.strip().split(" ")
     uniq_items = sorted({key: None for key in items}.keys())
     return uniq_items
@@ -107,7 +100,14 @@ def _symlinks(repo_ctx, paths):
         repo_ctx.symlink("/" + path, path)
 
 def _pkg_config_repository_impl(repo_ctx):
-    includes = _exec_pkg_config(repo_ctx, "--cflags-only-I")
+    includes = _exec_pkg_config(repo_ctx, ["--cflags-only-I"])
+
+    # If includes is empty, pkg-config will be re-executed with
+    # the --keep-system-cflags option added. Typically, -I/usr/include is
+    # returned, enabling bazel to recognize packages as valid even when
+    # pkg-config does not output cflags with standard options.
+    if includes[0] == "":
+        includes = _exec_pkg_config(repo_ctx, ["--cflags-only-I", "--keep-system-cflags"])
     includes = [item[len("-I/"):] for item in includes]
     _symlinks(repo_ctx, includes)
     data = {
@@ -116,13 +116,13 @@ def _pkg_config_repository_impl(repo_ctx):
         # https://github.com/bazelbuild/bazel/issues/23127
         "name": repo_ctx.attr.name.replace("~", "+").split("+")[-1],
         "hdrs": _make_strlist([item + "/**" for item in includes]),
-        "copts": _make_strlist(_exec_pkg_config(repo_ctx, "--cflags-only-other")),
+        "copts": _make_strlist(_exec_pkg_config(repo_ctx, ["--cflags-only-other"])),
         "includes": _make_strlist(includes),
-        "linkopts": _make_strlist(_exec_pkg_config(repo_ctx, "--libs-only-l")),
+        "linkopts": _make_strlist(_exec_pkg_config(repo_ctx, ["--libs-only-l"])),
     }
     build_file_data = BUILD_TEMPLATE.format(**data)
 
-    libexecdir = _exec_pkg_config(repo_ctx, "--variable=libexecdir")
+    libexecdir = _exec_pkg_config(repo_ctx, ["--variable=libexecdir"])
     if len(libexecdir) == 1:
         repo_ctx.symlink(libexecdir[0], "libexec")
         build_file_data += EXPORTS_FILES_TEMPLATE
--- a/src/bazel/pkg_config_repository.bzl
+++ b/src/bazel/pkg_config_repository.bzl
@@ -86,6 +86,13 @@ def _exec_pkg_config(repo_ctx, flag):
         print("pkg-config is not found")  # buildifier: disable=print
         return []
     result = repo_ctx.execute([binary, flag] + repo_ctx.attr.packages)
+
+    # If result.stdout.strip() is empty, pkg-config will be re-executed with
+    # the --keep-system-cflags option added. Typically, -I/usr/include is
+    # returned, enabling bazel to recognize packages as valid even when
+    # pkg-config does not output cflags with standard options.
+    if result.stdout.strip() == "":
+        result = repo_ctx.execute([binary, flag] + ["--keep-system-cflags"] + repo_ctx.attr.packages)
     items = result.stdout.strip().split(" ")
     uniq_items = sorted({key: None for key in items}.keys())
     return uniq_items


phoepsilonix:
Related https://github.com/fcitx/mozc/pull/68

hiroyuki-komatsu:
The root cause sounds that `_exec_pkg_config` returns `['']` when `pkg-config --cflags-only-I` returns an empty string. So how about updating `_exec_pkg_config` as follows instead of modifying `_pkg_config_repository_impl`?

```
def _exec_pkg_config(repo_ctx, flag):
    binary = repo_ctx.which("pkg-config")
    if not binary:
        # Using print is not recommended, but this will be a clue to debug build errors in
        # the case of pkg-config is not found.
        print("pkg-config is not found")  # buildifier: disable=print
        return []
    result = repo_ctx.execute([binary, flag] + repo_ctx.attr.packages)
    items = result.stdout.strip().split(" ")

+   # If items is empty, return an empty list.
+   if len(items) == 1 and items[0] == '':
+     return []

    uniq_items = sorted({key: None for key in items}.keys())
    return uniq_items
```

phoepsilonix:
With that change, @fcitx is still not recognized and cannot be found. If it is empty ([]), an error occurs. Return -I/usr/include. If possible, it is preferable to return --keep-system-cflags.
```
 ../scripts/build_fcitx_bazel 
Starting local Bazel server and connecting to it...
DEBUG: /home/phoepsilonix/.cache/bazel/_bazel_phoepsilonix/a422ec5eb6e7740277eb4ee8963a4986/external/rules_android_ndk~/rules.bzl:28:14: Either the ANDROID_NDK_HOME environment variable or the path attribute of android_ndk_repository must be set.
ERROR: Traceback (most recent call last):
	File "/home/phoepsilonix/.cache/bazel/_bazel_phoepsilonix/a422ec5eb6e7740277eb4ee8963a4986/external/_main~_repo_rules~fcitx/BUILD.bazel", line 8, column 16, in <toplevel>
		hdrs = glob([
Error in glob: pattern cannot be empty
ERROR: /home/phoepsilonix/.cache/bazel/_bazel_phoepsilonix/a422ec5eb6e7740277eb4ee8963a4986/external/_main~_repo_rules~fcitx/BUILD.bazel: no such target '@@_main~_repo_rules~fcitx//:fcitx': target 'fcitx' not declared in package '' defined by /home/phoepsilonix/.cache/bazel/_bazel_phoepsilonix/a422ec5eb6e7740277eb4ee8963a4986/external/_main~_repo_rules~fcitx/BUILD.bazel
ERROR: /home/phoepsilonix/work/aur/mozc-with-jp-dict/src/mozc/src/unix/fcitx/BUILD.bazel:89:15: no such target '@@_main~_repo_rules~fcitx//:fcitx': target 'fcitx' not declared in package '' defined by /home/phoepsilonix/.cache/bazel/_bazel_phoepsilonix/a422ec5eb6e7740277eb4ee8963a4986/external/_main~_repo_rules~fcitx/BUILD.bazel and referenced by '//unix/fcitx:fcitx-mozc.so'
ERROR: Analysis of target '//unix/fcitx:fcitx-mozc.so' failed; build aborted: Analysis failed
INFO: Elapsed time: 11.598s, Critical Path: 0.05s
INFO: 1 process: 1 internal.
ERROR: Build did NOT complete successfully
```
```
 cat /home/phoepsilonix/.cache/bazel/_bazel_phoepsilonix/a422ec5eb6e7740277eb4ee8963a4986/external/_main~_repo_rules~fcitx/BUILD.bazel 

package(
    default_visibility = ["//visibility:public"],
)

cc_library(
    name = "fcitx",
    hdrs = glob([
        ""
    ]),
    copts = [
        ""
    ],
    includes = [
        ""
    ],
    linkopts = [
        "-lfcitx-config",
        "-lfcitx-core",
        "-lfcitx-utils"
    ],
)
```

phoepsilonix:
```diff
diff --git a/src/bazel/pkg_config_repository.bzl b/src/bazel/pkg_config_repository.bzl
index 8d46e01..f60e9e9 100644
--- a/src/bazel/pkg_config_repository.bzl
+++ b/src/bazel/pkg_config_repository.bzl
@@ -87,6 +87,11 @@ def _exec_pkg_config(repo_ctx, flag):
         return []
     result = repo_ctx.execute([binary, flag] + repo_ctx.attr.packages)
     items = result.stdout.strip().split(" ")
+    # If items is empty, return with --keep-system-cflags.
+    if len(items) == 1 and items == "":
+        result = repo_ctx.execute([binary, flag] + ["--keep-system-cflags"] + repo_ctx.attr.packages)
+        items = result.stdout.strip().split(" ")
+
     uniq_items = sorted({key: None for key in items}.keys())
     return uniq_items
 
```

phoepsilonix:
```diff
diff --git a/src/bazel/pkg_config_repository.bzl b/src/bazel/pkg_config_repository.bzl
index 8d46e01..519a7ee 100644
--- a/src/bazel/pkg_config_repository.bzl
+++ b/src/bazel/pkg_config_repository.bzl
@@ -86,6 +86,9 @@ def _exec_pkg_config(repo_ctx, flag):
         print("pkg-config is not found")  # buildifier: disable=print
         return []
     result = repo_ctx.execute([binary, flag] + repo_ctx.attr.packages)
+    # If results.stdout is empty, return with --keep-system-cflags.
+    if result.stdout.strip() == "":
+        result = repo_ctx.execute([binary, flag] + ["--keep-system-cflags"] + repo_ctx.attr.packages)
     items = result.stdout.strip().split(" ")
     uniq_items = sorted({key: None for key in items}.keys())
     return uniq_items
```

phoepsilonix:
With this change, `@fcitx` is now recognized as a package.
```
cat /home/phoepsilonix/.cache/bazel/_bazel_phoepsilonix/a422ec5eb6e7740277eb4ee8963a4986/external/_main~_repo_rules~fcitx/BUILD.bazel 

package(
    default_visibility = ["//visibility:public"],
)

cc_library(
    name = "fcitx",
    hdrs = glob([
        "usr/include/**"
    ]),
    copts = [
        ""
    ],
    includes = [
        "usr/include"
    ],
    linkopts = [
        "-lfcitx-config",
        "-lfcitx-core",
        "-lfcitx-utils"
    ],
)

exports_files(glob(["libexec/*"]))
```

phoepsilonix:
I fixed the commit comment.

src/bazel/pkg_config_repository.bzl
```diff
@@ -86,6 +86,10 @@ def _exec_pkg_config(repo_ctx, flag):
         print("pkg-config is not found")  # buildifier: disable=print
         return []
     result = repo_ctx.execute([binary, flag] + repo_ctx.attr.packages)
+    # If result.stdout.strip() is empty, pkg-config will be re-executed with the --keep-system-cflags option added.
```
yukawa, line 89: Can you run [`buildifier`](https://github.com/bazelbuild/buildtools/blob/main/buildifier/README.md) to apply the standard format?

Otherwise, you would see CI error as follow.
https://github.com/phoepsilonix/mozc/actions/runs/11536533465/job/32112886990

src/bazel/pkg_config_repository.bzl
```diff
@@ -86,6 +86,10 @@ def _exec_pkg_config(repo_ctx, flag):
...
+    # Typically, -I/usr/include is returned, enabling Bazel to recognize packages that do not output cflags through pkg-config.
+    if result.stdout.strip() == "":
+        result = repo_ctx.execute([binary, flag] + ["--keep-system-cflags"] + repo_ctx.attr.packages)
```
yukawa, line 92: Can you wrap the comment so that each line is within 80 column?

While line length limit is [not strictly enforced in starlark files](https://bazel.build/rules/bzl-style#line-length), I personally feel it's convenient when reviewing code with side-by-side diff if each line is within 80 column.

phoepsilonix: I see.
Wait a moments

src/bazel/pkg_config_repository.bzl
phoepsilonix: I see.

phoepsilonix:
```sh
go install github.com/bazelbuild/buildtools/buildifier@latest
$(go env GOPATH)/bin/buildifier --mode=diff --diff_command="diff -u" -r .
```
output is nothing.

I wrapped comments and commit logs at 80 characters.

src/bazel/pkg_config_repository.bzl
phoepsilonix:
```sh
go install github.com/bazelbuild/buildtools/buildifier@latest
$(go env GOPATH)/bin/buildifier --mode=diff --diff_command="diff -u" -r .
```
output is nothing.

src/bazel/pkg_config_repository.bzl
phoepsilonix: I wrapped comments and commit logs at 80 characters.

phoepsilonix:
I made some minor edits to the English in the commit log.

phoepsilonix:
I thought it was unnecessary because it would result in a duplicate check, so I disabled it, but I enabled CI actions for each OS in the fork repository. I'll enable it when I create a PR, etc.

wengxt:
I don't think that pull in the whole usr/include is an expected behavior.

The purpose of repository is to pull in only necessarily files.

I'd suggest an alternative solution, that simply make pkg_config_repository accept a manual override on the include directory, or to make it derived from a certain pkg-config variable.

```
pkg_config_repository(
    name = ..
    includepath = [
         "$(includedir}/subdir1,
         "$(includedir}/subdir2,
         "$(includedir}/subdir3,
     ]
)
```

However, I'm fine if this is not fixed and keep it as is. I'm ok to use new_local_repository and do some manual config instead.

phoepsilonix:
Even if there are files in /usr/include/fcitx{,-config,-gclient,-utils}, the header inclusion will include folders such as <fcitx/ime.h>. Therefore, we can only expect pkg-config to output with -I/usr/include.

The template is based on globbing, so if we were to make a change like that, we would have to change that part first, right?
If an includepath is specified, it seems necessary to do things like switching the output of hdrs, etc.
```
    hdrs = glob([
        {hdrs}
    ]),
```
I certainly agree that there is an option to not use pkg-config.
As for the headers, it would be good to write settings to include the files in these folders(/usr/include/fcitx{,-config,-gclient,-utils}).
It would probably be good to include the libraries /usr/lib/libfcitx-*,/usr/lib/fcitx/.

Adding -keep-system-cflags is only necessary if cflags is not output, so I think it will have little impact on other packages. And since it is entirely possible for cflags to not be output by pkg-config, I think it is necessary to add a response for that case.

phoepsilonix:
It seems that CI for each OS, including lint-check, was completed successfully.

phoepsilonix:
```diff
diff --git a/src/MODULE.bazel b/src/MODULE.bazel
index df10ba2..8042d51 100644
--- a/src/MODULE.bazel
+++ b/src/MODULE.bazel
@@ -90,7 +90,6 @@ android_ndk_repository_extension = use_extension(
 )
 use_repo(android_ndk_repository_extension, "androidndk")
 
-register_toolchains("@androidndk//:all")
 
 new_local_repository = use_repo_rule(
     "@bazel_tools//tools/build_defs/repo:local.bzl",
@@ -130,7 +129,13 @@ pkg_config_repository(
 # Fcitx
 pkg_config_repository(
     name = "fcitx",
-    packages = ["fcitx"],
+    hdrs = [
+        "usr/include/fcitx",
+        "usr/include/fcitx-config",
+        "usr/include/fcitx-gclient",
+        "usr/include/fcitx-utils",
+    ],
+    packages = [ "fcitx" ],
 )
 
 # Fcitx 5
diff --git a/src/bazel/pkg_config_repository.bzl b/src/bazel/pkg_config_repository.bzl
index 8d46e01..7e793e6 100644
--- a/src/bazel/pkg_config_repository.bzl
+++ b/src/bazel/pkg_config_repository.bzl
@@ -86,6 +86,11 @@ def _exec_pkg_config(repo_ctx, flag):
         print("pkg-config is not found")  # buildifier: disable=print
         return []
     result = repo_ctx.execute([binary, flag] + repo_ctx.attr.packages)
+
+    # If result.stdout.strip() is empty, pkg-config will be re-executed with the --keep-system-cflags option added.
+    # Typically, -I/usr/include is returned, enabling Bazel to recognize packages that do not output cflags through pkg-config.
+    if result.stdout.strip() == "":
+        result = repo_ctx.execute([binary, flag] + ["--keep-system-cflags"] + repo_ctx.attr.packages)
     items = result.stdout.strip().split(" ")
     uniq_items = sorted({key: None for key in items}.keys())
     return uniq_items
@@ -100,8 +105,11 @@ def _symlinks(repo_ctx, paths):
         repo_ctx.symlink("/" + path, path)
 
 def _pkg_config_repository_impl(repo_ctx):
-    includes = _exec_pkg_config(repo_ctx, "--cflags-only-I")
-    includes = [item[len("-I/"):] for item in includes]
+    if len(repo_ctx.attr.hdrs) != 0:
+        includes = repo_ctx.attr.hdrs
+    else:
+        includes = _exec_pkg_config(repo_ctx, "--cflags-only-I")
+        includes = [item[len("-I/"):] for item in includes]
     _symlinks(repo_ctx, includes)
     data = {
         # In bzlmod, repo_ctx.attr.name has a prefix like "_main~_repo_rules~ibus".
@@ -128,5 +136,6 @@ pkg_config_repository = repository_rule(
     local = True,
     attrs = {
         "packages": attr.string_list(),
+        "hdrs": attr.string_list(),
     },
 )
```

phoepsilonix:
If this feature addition is better, I will include it in this PR. What do you think?

phoepsilonix:
@yukawa @wengxt @hiroyuki-komatsu
If this feature addition is better, I will include it in this PR. What do you all think?

phoepsilonix:
For now, I'll add a commit for the new features. I'll wait for the CI check.

phoepsilonix:
It seems that CI for each OS, including lint-checks, was completed successfully.

yukawa:
> However, I'm fine if this is not fixed and keep it as is. I'm ok to use new_local_repository and do some manual config instead.

Yeah, I agree that just using `new_local_repository` would give the biggest flexibility to the actual `@fcitx` repository owner. In that way you can keep improving things without upstreaming changes to google/mozc repository.

So my suggests are:

 * use `new_local_repository` for now to address immediate needs to make `fcitx-mozc` buildable with Bazel.
 * revisit the issue later (maybe 1 year or so) if it's still relevant. Hopefully we would have better understandings on how to deal with this kind of requirements with Bazel.

phoepsilonix:
Of course, we can also solve the problem in the following way.

```diff
diff --git a/src/MODULE.bazel b/src/MODULE.bazel
index df10ba2..9522b56 100644
--- a/src/MODULE.bazel
+++ b/src/MODULE.bazel
@@ -128,9 +128,10 @@ pkg_config_repository(
 )
 
 # Fcitx
-pkg_config_repository(
+new_local_repository(
     name = "fcitx",
-    packages = ["fcitx"],
+    path = "/usr",
+    build_file = "unix/fcitx/BUILD.fcitx.bazel",
 )
 
 # Fcitx 5
diff --git a/src/bazel/pkg_config_repository.bzl b/src/bazel/pkg_config_repository.bzl
index 8d46e01..b750170 100644
--- a/src/bazel/pkg_config_repository.bzl
+++ b/src/bazel/pkg_config_repository.bzl
@@ -86,6 +86,11 @@ def _exec_pkg_config(repo_ctx, flag):
         print("pkg-config is not found")  # buildifier: disable=print
         return []
     result = repo_ctx.execute([binary, flag] + repo_ctx.attr.packages)
+
+    # If result.stdout.strip() is empty, pkg-config will be re-executed with the --keep-system-cflags option added.
+    # Typically, -I/usr/include is returned, enabling Bazel to recognize packages that do not output cflags through pkg-config.
+    if result.stdout.strip() == "":
+        result = repo_ctx.execute([binary, flag] + ["--keep-system-cflags"] + repo_ctx.attr.packages)
     items = result.stdout.strip().split(" ")
     uniq_items = sorted({key: None for key in items}.keys())
     return uniq_items
diff --git a/src/unix/fcitx/BUILD.fcitx.bazel b/src/unix/fcitx/BUILD.fcitx.bazel
new file mode 100644
index 0000000..3b0a229
--- /dev/null
+++ b/src/unix/fcitx/BUILD.fcitx.bazel
@@ -0,0 +1,29 @@
+package(
+    default_visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "fcitx",
+    hdrs = glob([
+        "usr/include/fcitx/**",
+        "usr/include/fcitx-config/**",
+        "usr/include/fcitx-gclient/**",
+        "usr/include/fcitx-utils/**"
+    ]),
+    copts = [
+        ""
+    ],
+    includes = [
+        "usr/include/fcitx",
+        "usr/include/fcitx-config",
+        "usr/include/fcitx-gclient",
+        "usr/include/fcitx-utils"
+    ],
+    linkopts = [
+        "-lfcitx-config",
+        "-lfcitx-core",
+        "-lfcitx-utils"
+    ],
+)
+
+exports_files(glob(["libexec/*"]))
```

src/bazel/pkg_config_repository.bzl
```diff
@@ -86,6 +86,13 @@ def _exec_pkg_config(repo_ctx, flag):
         print("pkg-config is not found")  # buildifier: disable=print
         return []
     result = repo_ctx.execute([binary, flag] + repo_ctx.attr.packages)
+
+    # If result.stdout.strip() is empty, pkg-config will be re-executed with
+    # the --keep-system-cflags option added. Typically, -I/usr/include is
+    # returned, enabling bazel to recognize packages as valid even when
+    # pkg-config does not output cflags with standard options.
+    if result.stdout.strip() == "":
+        result = repo_ctx.execute([binary, flag] + ["--keep-system-cflags"] + repo_ctx.attr.packages)
```
hiroyuki-komatsu, line 95: Should `--keep-system-cflags` be applied only when the output of `--cflags-only-I` is empty?  If yes, please move this logic after `_exec_pkg_config(repo_ctx, "--cflags-only-I")` in `_pkg_config_repository_impl`.

It's fine to modify `flag` to `flags` that takes a list of string.

phoepsilonix: How about a modification like this?

phoepsilonix: It seems that the CI check has also passed.

hiroyuki-komatsu: Thank you for the update.
I'm running the internal process now. Please keep this PR as-is.

hiroyuki-komatsu:
We have merged this PR.

Issue #1121, opened 2024-11-13
Use `QT_HOST_PATH` to cross compile Qt 6.8


yukawa:
## Description
As seen in the following build result, Qt 6.8.0 looks to be requiring us to use `QT_HOST_PATH` to build Qt for CPU architecture other than the host environment.

 * https://github.com/google/mozc/actions/runs/11793782247

```
Overriding option 'libpng' with 'qt' (was: 'undefined')
[...]

PR #1125, opened 2024-11-19
Improve error messages in case MSVC for ARM64 is not explicitly installed

yukawa:
## Description

This is a follow up commit to my previous commit (a0fed621e7ae4fb99629631654e2a65b942c653b), which made build_qt.py compatible with ARM64 host and/or target.

One remaining pitfall is that ARM64 C++ compiler needs to be explicitly selected when installing Visual Studio on the x64 environment.

With this commit at least you can see a clear message that the compiler is missing when you try to build Qt for ARM64 without installing the required optional component.

## Issue IDs

 * https://github.com/google/mozc/issues/1121

 - OS: Windows 11 23H2 (x64)
 - Steps:
   1. Check out Mozc
   2. `python .\build_tools\update_deps.py`
   3. `python build_tools/build_qt.py --release --confirm_license --target_arch=arm64`
   4. You can see a clear message when C++ compiler for ARM64 is not explicitly installed.

### DIFF
--- a/src/build_tools/vs_util.py
+++ b/src/build_tools/vs_util.py
@@ -37,10 +37,14 @@ import sys
 from typing import Union
 
 
-def get_vcvarsall(path_hint: Union[str, None] = None) -> pathlib.Path:
+def get_vcvarsall(
+  arch: str,
+  path_hint: Union[str, None] = None
+) -> pathlib.Path:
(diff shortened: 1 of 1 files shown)


hiroyuki-komatsu:
We have merged your PR.

PR #1128, opened 2024-11-21
Disallow Mozc to be installed on ARM64 Windows

### LINKED ISSUE #1127: Disallow Mozc to be installed into Windows ARM64 machines
## Description
While supporting ARM64 Window environment has been in our long term roadmap, it is supposed to require some non trivial amount of code changes. Until it becomes ready, let's update `Mozc64.msi` so that we can prevent it from being installed into Windows ARM64 machines.

## Steps to reproduce
1. Build Mozc
2. Try to install `Mozc64.msi` on an ARM64 Windows machine.

## Expected behavior
The installation fails.

## Actual behavior
The installation succeeds.

## Version or commit-id
df1bbd91020a8c2b06eadce5046e46351135de21

## Environment
 - OS: Windows 11 23H2 (arm64)

yukawa:
## Description
Before Mozc becomes available to ARM64 processes in Windows, let's disallow `Mozc64.msi` to be installed into ARM64 Windows machines to avoid unnecessary user confusions.

Note that this commit only blocks the initial installation. There is no behavior change for those who have already installed Mozc into their ARM64 Windows environments. This means that both the following operations continue to be allowed.

 * Upgrading from a previous version of Mozc on an ARM64 Windows environment
 * Uninstalling Mozc on an ARM64 Windows environment

Closes #1127.

## Issue IDs

 * https://github.com/google/mozc/issues/1127

 - OS: Windows 11 23H2
 - Steps:
   1. Try to install `Mozc64.msi`
   2. It fails with a dialog message with "ARM64 環境へのインストールは未対応です。"

### DIFF
--- a/src/win32/installer/installer_64bit.wxs
+++ b/src/win32/installer/installer_64bit.wxs
@@ -63,6 +63,11 @@
     <Launch Condition="Privileged" Message="Google 日本語入力をインストールするには管理者権限が必要です。" />
     <Media Id="1" Cabinet="GoogleJapaneseInput.cab" EmbedCab="yes" CompressionLevel="high" />
 
+    <Property Id="PROCESSOR_ARCHITECTURE">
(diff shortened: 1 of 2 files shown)


hiroyuki-komatsu:
We have merged your PR.

Issue #1130, opened 2024-12-03
ARM64 support for Window


yukawa:
## Description
This follows up to the following issue.

 * https://github.com/google/mozc/issues/1127

The goal here is to make it possible to use Mozc in Windows ARM64 environment at the minimal implementation and maintenance cost.

Supporting ARM64 environment to build Mozc is a non-goal of this task. It will be handled in the following separate task.

 * https://github.com/google/mozc/issues/1296

## Version or commit-id
006ed69bf545548a8a3596b13f58cb22cf3d8a2f

## Environment
 - OS: Windows 11 24H2

PR #1132, opened 2024-12-03
Remove platform macros from `tip_text_service.cc`

yukawa:
## Description
This commit removes the dependency on platform macros such as `_M_X64` and `_M_IX86` from `tip_text_service.cc`.

There must be no change in the final artifacts.

This is a preparation to build Mozc for ARM64 (#1130).

## Issue IDs

 - OS: Windows 11 23H2
 - Steps:
   1. Build and install `Mozc64.msi`
   2. Confirm you can type with Mozc on both 32-bit app and 64-bit app.

### DIFF
--- a/src/win32/tip/tip_text_service.cc
+++ b/src/win32/tip/tip_text_service.cc
@@ -36,6 +36,8 @@
 #include <wil/com.h>
 #include <windows.h>
 
+#include <algorithm>
+#include <bit>
 #include <cstddef>
 #include <cstdint>
 #include <memory>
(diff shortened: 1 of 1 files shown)


src/win32/tip/tip_text_service.cc
```diff
@@ -36,8 +36,10 @@
 #include <wil/com.h>
 #include <windows.h>
 
+#include <bit>
 #include <cstddef>
 #include <cstdint>
+#include <functional>
```
hiroyuki-komatsu, line 42: `#include <algorithm>` instead?

yukawa: Ah, yes. Fixed.

hiroyuki-komatsu:
We have merged your PR.

Issue #1002, opened 2024-08-15
Migration of Bazel dependency system to Bzlmod.


hiroyuki-komatsu:
Bzlmod is the new dependency system of Bazel. We need to migrate the current system (i.e. WORKSPACE) to Bzlmod since the current system will be deprecated.

https://bazel.build/external/overview#bzlmod

Once we complete the migration, we will remove the current system with WORKSPACE soon. This should not change anything for users, unless you use Mozc as a third party library in your Bazel WORKSPACE rules.

Please let us know if you have concerns about our migration plan.

hiroyuki-komatsu:
We have completed the migration from WORKSPACE to Bzlmod.
We will remove WORKSPACE.bazel in future.

Note,
* Bazel 8 will disable WORKSPACE by default.
* Bazel 9 will remove WORKSPACE.

https://bazel.build/about/roadmap

PR #1141, opened 2024-12-16
Remove `rules_java` from `MODULE.bazel`

yukawa:
## Description
This follows up to our previous commit (699c2ce5b7d92476199e9a6ba1d83414c3558016), which introduced the dependency on `rules_java` as an attempt to support Android build with bzlmod (#1002).

Luckily a subsequent commit (6c61c4f29a966264784d608584000a56596f3ed9) had effectively removed the the dependency on `rules_java`, but `MODULE.bazel` has not been reflected yet. Let's also update `MODULE.bazel` for simplicity and maintainability.

There must be no difference in the final artifacts.

## Issue IDs

 * https://github.com/google/mozc/issues/1002

### DIFF
--- a/src/MODULE.bazel
+++ b/src/MODULE.bazel
@@ -65,13 +65,6 @@ bazel_dep(
     repo_name = "build_bazel_apple_support",
 )
 
-# Java rules (7.9.0 2024-08-13)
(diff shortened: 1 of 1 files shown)


hiroyuki-komatsu:
We have merged your PR.

PR #1159, opened 2025-01-08
Gracefully handle empty output from `vswhere.exe`

yukawa:
## Description
This is a follow up commit to my previous commit (ace314567109cbc30dc336fb27844dacff782dd9), which started using 'vswhere.exe' to locate `vcvarsall.bat` as discussed in #1057.

One thing I overlooked is that `vswhere.exe` could return an empty result even when the exitcode is `ERROR_SUCCESS`. With this commit such a case will be gracefully handled.

This is also a preparation to support ARM64 build on Windows (#1130).

## Issue IDs

 * https://github.com/google/mozc/pull/1057

 - OS: Windows 11
 - Steps:
   1. You can still build Mozc with both GYP and Bazel.


hiroyuki-komatsu:
We have merged your PR.

PR #1162, opened 2025-01-19
Simplify NDK setup in `Dockerfile`

yukawa:
## Description
This commit aims to simplify how Android NDK is set up in `Dockerfile` without changing any final artifact.

With this commit Android NDK will be directly downloaded then deployed just by unzipping the archive. As a result we no longer need to install `openjdk-17-jdk` package just to run Android SDK setup.

This commit also makes it clear that the build environment for Android requires Android NDK only. Android SDK and `ANDROID_HOME` environment variable are not necessary.

There must be no difference in the final artifact.

## Issue IDs
N/A

 - OS: Ubuntu 24.04.1
 - Steps:
   1. You can build Mozc for Android inside Docker container.


hiroyuki-komatsu:
We have merged your PR.

PR #1199, opened 2025-02-26
Extract out Linux build instructions

### LINKED ISSUE #1181: Remove the dependency on Docker from Android/Linux build instructions
## Summary
While our `Dockerfile` has been helping us defining a reference build environment, it might be time to just simply explain how to build `libmozc.so` and binaries for Linux desktop without using `Dockerfile`

### Context
When we first introduced `Dockerfile` for Mozc (d7a7e610c14138982ab3faa40b6c841cf96ab0b1) more than 10 years ago, things were much more complicated.
 * Mozc supported NaCl and Android app on top of GYP.
   * Supporting NaCl required us to keep track of which version of NaCl SDK to be used.
 * There was no GitHub Actions. We had already relied on Travis CI though.

Now that we switched to Bazel and GitHub Actions, it could be acceptable for us to just explain the main building command and ask developers to follow our GitHub Actions rules for details for Linux/Android executables. This is actually what we are currently doing for macOS and Windows.

### Migration Plan

1. Split out the build instructions for `libmozc.so` into a separate file (e.g. `build_mozc_for_android.md`) from the existing [`build_mozc_in_docker.md`](https://github.com/google/mozc/blob/f8bb1996239800e9507a8cc119cddc7239f6984d/docs/build_mozc_in_docker.md). In this step, we can also make it clear that building `libmozc.so` is indeed now possible not only in Linux environment but also in macOS environment. We can also update the existing `update_deps.py` so that the expected version of Android NDK can be downloaded for the running environment.
2. Split out the build instructions for Linux desktop into a separate file (e.g. `build_mozc_for_linux.md`) from the existing [`build_mozc_in_docker.md`](https://github.com/google/mozc/blob/f8bb1996239800e9507a8cc119cddc7239f6984d/docs/build_mozc_in_docker.md). For existing web pages that currently refer to `build_mozc_in_docker.md`, we can keep `build_mozc_in_docker.md` to explain that the content has moved to a new file.

## Version or commit-id
f8bb1996239800e9507a8cc119cddc7239f6984d

yukawa:
## Description
As the final step towards removing the dependency on Docker from build instructions, this commit introduces `build_mozc_for_linux.md` as the successor of long-lived `build_mozc_in_docker.md`.

This commit also declares that our `Dockerfile` is now deprecated.

Closes #1181.

## Issue IDs

 * https://github.com/google/mozc/issues/1181


hiroyuki-komatsu:
We have merged your PR.

Issue #1179, opened 2025-02-09
See if we can use `clang-cl` to build Mozc for Windows with Bazel


yukawa:
## Description
Spun off from the following issues.

 * https://github.com/google/mozc/issues/1177
 * https://github.com/protocolbuffers/protobuf/issues/20085

The build failure in protobuf with bazel and `cl.exe` due to path length limit has been an issue in Mozc as well.
As you can see below, we are now passing `--output_base=C:/x` to shorten the absolute path.
https://github.com/google/mozc/blob/3baa8d865d8f7a99f30b5ba34a4419dbfdebb012/src/windows.bazelrc#L10-L15

If protobuf team is even giving up this workaround, maybe it's time for us to switch to `clang-cl.exe` as well.

Let's see what tasks need to be done to switch to `clang-cl.exe` in Mozc.

## Version or commit-id
3baa8d865d8f7a99f30b5ba34a4419dbfdebb012

## Environment
 - OS: Windows 24H2

yukawa:
One of the blocker is the fact that `clang-cl` configured with `rules_cc` does not support to generate 32-bit executables right now.

 * https://github.com/bazelbuild/rules_cc/issues/358

PR #1201, opened 2025-02-27
Introduce a stub bazel wrapper for Windows

yukawa:
## Description
This is a preparation before starting using `clang-cl` to build Mozc for for Windows with Bazel (#1179).

In order to make the build process reliable and reproducible, we would like to download a specific version of LLVM with 'update_deps.py' then deploy it under `third_party/llvm/` and let `rules_cc` use it by specifying `BAZEL_LLVM` environment variable.

 * https://bazel.build/configure/windows#clang

To automate this process, one idea is to use `tools/bazel` wrapper script.

 * https://github.com/bazelbuild/bazelisk/blob/master/README.md#toolsbazel

Luckily the wrapper script directory is also customizable since bazelisk v1.21.0.

 * https://github.com/bazelbuild/bazelisk/pull/567

With above this commit introduces a stub `bazel.bat` so that we can use it to set various environment variables for Windows bazel build in subsequent commits. At this moment there must be no difference in the final artifacts. There should be no impact on non-Windows bazel build as `bazel.bat` is searched only on Windows.

## Issue IDs

 * https://github.com/google/mozc/issues/1179

 - OS: Windows 11 24H2
 - Steps:
   1. ` bazelisk --bazelrc=windows.bazelrc build package --config oss_windows --config release_build`
   2. Confirm the build succeeds.


yukawa:
Rebased to address merge conflict.

src/bazel/bazel_wrapper/README.md
```diff
@@ -0,0 +1,13 @@
+# bazel_wrapper
+
+This directory can be used to place special wrapper scripts for bazelisk.
+
+ * [tools/bazel - Bazelisk](https://github.com/bazelbuild/bazelisk/blob/master/README.md#toolsbazel)
```
hiroyuki-komatsu, line 5: Please update the format as follows.

```
*   [tools/bazel - Bazelisk](https://github.com/bazelbuild/bazelisk/blob/master/README.md#toolsbazel)

By default `bazel` and `bazelisk` uses `tools/` as the bazel wrapper directory,
but `bazelisk` allows us to customize it with `BAZELISK_WRAPPER_DIRECTORY` in
[`src/.bazeliskrc`](../../.bazeliskrc) since bazelisk v1.21.0.

*   [Allow overriding tools/bazel path with BAZELISK_WRAPPER_DIRECTORY](https://github.com/bazelbuild/bazelisk/pull/567)

## Contents

*   [bazel.bat](./bazel.bat)
```

yukawa: Done.

hiroyuki-komatsu:
We have merged your PR.

Issue #1108, opened 2024-11-01
The symbol file name embedded in `mozc_tip32.dll` built with Bazel is `mozc_tip.pdb` rather than `mozc_tip32.dll.pdb`


yukawa:
## Description
This follows up to the following task, where we started deploying `mozc_tip32.dll.pdb` to users' environments.

 * https://github.com/google/mozc/issues/1081
 * https://github.com/google/mozc/pull/1082

As noted in the commit 0377ddd0b617c824b93d17e87bab028acc0ef3e7, the symbol file name embedded in `mozc_tip32.dll` built with Bazel is `mozc_tip.pdb` rather than `mozc_tip32.dll.pdb`. Let's fix this so that we can fully switch to Bazel build to deprecate GYP build.

## Steps to reproduce
1. `bazel --bazelrc=windows.bazelrc build --config oss_windows --config release_build //win32/installer:mozc_tip32`
2. `cp bazel-bin\win32\installer\mozc_tip32.dll .`
3. `dumpbin /headers mozc_tip32.dll | findstr cv`

## Expected behavior
You would see `mozc_tip32.dll.pdb` in the output.

## Actual behavior
You see `mozc_tip.pdb` in the output.

## Version or commit-id
ad3bee0547e3cb5c0aa0b94437acccafdc6cfaad

## Environment
 - OS: Windows 11 23H2

PR #1207, opened 2025-03-12
Preserve features in `mozc_win32_cc_prod_binary`

### LINKED ISSUE #1206: `--features` command line option can be ignored in Windows bazel build
## Description
Even if you explicitly specify [`--features`](https://bazel.build/reference/command-line-reference#flag--features) command line option, such a feature can be completely ignored if the build target is using [`mozc_win32_cc_prod_binary`](https://github.com/google/mozc/blob/e453da9454530e714e9e0ce542f3fc7684dcebe9/src/build_defs.bzl#L279C5-L279C30).

https://github.com/google/mozc/blob/e453da9454530e714e9e0ce542f3fc7684dcebe9/src/renderer/win32/BUILD.bazel#L48-L69

`treat_warnings_as_errors` defined in `rules_cc` is an example.

https://github.com/bazelbuild/rules_cc/blob/07c5535812e262dd21fdcf7af8fd60f73148d0bc/cc/private/toolchain/windows_cc_toolchain_config.bzl#L946-L954
```starlark
        treat_warnings_as_errors_feature = feature(
            name = "treat_warnings_as_errors",
            flag_sets = [
                flag_set(
                    actions = [ACTION_NAMES.c_compile, ACTION_NAMES.cpp_compile] + all_link_actions,
                    flag_groups = [flag_group(flags = ["/WX"])],
                ),
            ],
        )
```

## Steps to reproduce
```
bazelisk --bazelrc=windows.bazelrc build package --config oss_windows --config release_build --features=treat_warnings_as_errors
```

## Expected behavior
The build command fails with `error C2220` due to some existing warnings.

## Actual behavior
The build command succeeds.

## Version or commit-id
e453da9454530e714e9e0ce542f3fc7684dcebe9

## Environment
 - OS: Windows 11 24H2

## Additional context
The root cause is that `_win_executable_transition` does not preserve the existing `features`. It always replaces the entire `features` set with `generate_pdb_file` (and `static_link_msvcrt` if `static_crt` is set).

https://github.com/google/mozc/blob/e453da9454530e714e9e0ce542f3fc7684dcebe9/src/build_defs.bzl#L208-L226

We should be able to preserve the existing features there.

yukawa:
## Description
This follows up to our previous commits (bc546b239bccff494773104e108ba08fcff40d5b)(51b4cd065d78aac6bec9a6152b9d96de00ec9c76), where we start using [transitions](https://bazel.build/rules/lib/builtins/transition) for `//command_line_option:features`.

One thing we have overlooked is that our previous implementation does not take existing `--features` command line options into account. As a result, such existing `--features` command line options will be always discarded and have no effect.

With this commit `mozc_win32_cc_prod_binary` starts preserving existing `--features` command line options when creating a transition.

Closes #1206.

## Issue IDs

 * https://github.com/google/mozc/issues/1108
 * https://github.com/google/mozc/issues/1206

 - OS: Windows 11 24H2
 - Steps:
   1. `bazelisk --bazelrc=windows.bazelrc build package --config oss_windows --config release_build --features=treat_warnings_as_errors`
   2. Confirm the build fails with `error C2220` due to existing warnings.

### DIFF
--- a/src/build_defs.bzl
+++ b/src/build_defs.bzl
@@ -205,12 +205,45 @@ register_extension_info(
     label_regex_for_dep = "{extension_name}",
 )
 
+def _append_if_absent(target_list, item):
+    """Append the given item only when it is absent in the target.
+
+    Args:
+      target_list: a String list to be checked.
+      item: a String item that is to be made sure to exist in the list.
+    Returns:
+      list: target_list itself or a new list, where the given item exists.
+    """
+    if item in target_list:
+        return target_list
+    return target_list + [item]
+
+def _remove_if_present(target_list, item):
+    """Remove the given item if exists.
+
+    Args:
+      target_list: a String list to be checked.
+      item: a String item that is to be made sure to not exist in the list.
+    Returns:
+      list: target_list itself or a new list, where the given item does not
+            exist.
+    """
+    return [x for x in target_list if x != item]
+
 def _win_executable_transition_impl(
-        settings,  # @unused
+        settings,
         attr):
-    features = ["generate_pdb_file"]
+    features = settings["//command_line_option:features"]
+
+    features = _append_if_absent(features, "generate_pdb_file")
+
     if attr.static_crt:
-        features.append("static_link_msvcrt")
+        features = _remove_if_present(features, "dynamic_link_msvcrt")
+        features = _append_if_absent(features, "static_link_msvcrt")
+    else:
+        features = _remove_if_present(features, "static_link_msvcrt")
+        features = _append_if_absent(features, "dynamic_link_msvcrt")
+
     return {
         "//command_line_option:features": features,
         "//command_line_option:platforms": [attr.platform],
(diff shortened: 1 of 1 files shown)


src/build_defs.bzl
```diff
@@ -205,20 +205,53 @@ register_extension_info(
     label_regex_for_dep = "{extension_name}",
 )
 
+def _append_if_absent(target_list, item):
+    """Append the given item only when it is absent in the target.
+    Args:
```
hiroyuki-komatsu, line 210: Please add one blank line before "Args:".

```
    """Append the given item only when it is absent in the target.

    Args:
```

src/build_defs.bzl
```diff
@@ -205,20 +205,53 @@ register_extension_info(
...
+    Returns:
+      list: target_list itself or a new list, where the given item exists.
+    """
+    if item in target_list:
+        return target_list
+    return target_list + [item]
+
+def _remove_if_present(target_list, item):
+    """Remove the given item if exists.
+    Args:
```
hiroyuki-komatsu, line 222: Here too.

src/build_defs.bzl
yukawa: Done.

src/build_defs.bzl
yukawa: Done.

PR #1198, opened 2025-02-26
Start using `clang-cl` for Windows Bazel build

yukawa:
## Description
With this commit we start using `clang-cl` to build Mozc for Windows with Bazel.

While the original motivation of switching to `clang-cl` is that protobuf is planning to stop supporting `cl.exe` with bazel build in protobuf v34, there is also an on-going build breakage that happens when combining protobuf and abseil-cpp 20250127.0, which is not yet fixed.
 
 * https://github.com/protocolbuffers/protobuf/issues/20085
 * https://github.com/protocolbuffers/protobuf/issues/20331

The latter one is more critical for us because more and more bzlmod dependencies start requiring abseil-cpp 20250127.0 or higher.

There must be no impact on Windows GYP build.

Closes #1179.

## Issue IDs

 * https://github.com/google/mozc/issues/1179

## Steps to test new behaviors (if any)
A clear and concise description about how to verify new behaviors (if any).
 - OS: Windows 11 24H2
 - Steps:
   1. `bazelisk --bazelrc=windows.bazelrc build package --config oss_windows --config release_build`
   2. Confirm you can build Mozc for Windows with Bazel. Also confirm `clang-cl.exe` is running during the build.

PR #1217, opened 2025-03-16
Make `windows.bazelrc` no-op before deprecating it

yukawa:
## Description
This follows up to my previous commit (9adf5cedd493084d2663a9b07f7808a9afa3bea9), which introduced `src/windows.bazelrc` to work around the path length limitation in cl.exe.

Now that Windows Bazel build has switched to clang-cl (#1198), the above workaround can also be removed. As the first step, this commit moves the remaining lines into `.bazelrc`. The actual file removal will happen in a subsequent commit.

No user-observable behavior change is intended in the final artifacts.

## Issue IDs

 * https://github.com/google/mozc/issues/948
 * https://github.com/google/mozc/issues/1179

 - OS: Windows 11 24H2
 - Steps:
```
bazelisk build --config oss_windows --config release_build package
```

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -123,7 +123,7 @@ jobs:
         env:
           ANDROID_NDK_HOME: ""
         run: |
-          bazelisk --bazelrc=windows.bazelrc build --config oss_windows --config release_build package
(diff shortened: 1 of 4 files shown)


hiroyuki-komatsu:
We have merged your PRs.

Issue #1223, opened 2025-03-19
Make `bazelisk test ...` run and pass on Windows


yukawa:
## Description
Currently there are multiple build and test failures when trying to run `bazelisk test` on Windows as follows.
```
bazelisk test ... --config oss_windows -c dbg --build_tests_only
```

 * Fix all the build failures.
 * Fix all the test failures.
 * Set up GitHub Actions to continuously run these unit tests.

## Steps to reproduce
1. `bazelisk test ... --config oss_windows -c dbg --build_tests_only`

## Expected behavior
All the tests pass.

## Actual behavior
See multiple build failures. Even after working around such build failures, we would still see roughly 20 test failures.

## Version or commit-id
96a8a29fb85a3bf433e08376e3546025857e931a

## Environment
 - OS: Windows 11 24H2

PR #1227, opened 2025-03-19
Fix the Windows build of `//unix/ibus:gen_main_h`

yukawa:
## Description
As part of running unit tests with Bazel on Windows (#1223), this commit fixes an existing build failure of `//unix/ibus:gen_main_h`, which is required by `//unix/ibus:ibus_config_test`.

Perhaps test targets under `//unix/ibus/...` should have been annotated with `target_compatible_with` so that they can be excluded on non-Linux desktop builds, but at this point just fixing `//unix/ibus:gen_main_h` on Windows is easier.

No user-visible behavior change is intended in the final artifacts.

## Issue IDs

 * https://github.com/google/mozc/issues/1223

 - OS: Windows 11 24H2
 - Steps:
   1. .Confirm `bazelisk build //unix/ibus:gen_main_h --config oss_windows` succeeds

### DIFF
--- a/src/unix/ibus/gen_mozc_xml.py
+++ b/src/unix/ibus/gen_mozc_xml.py
@@ -174,10 +174,10 @@ def OutputCpp(component, engine_common, engines):
     OutputCppVariable('Engine', key, value)
   print('const size_t kEngineArrayLen = %s;' % len(engines))
   print('const char kEnginesXml[] = R"#(', end='')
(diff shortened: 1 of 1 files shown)


yukawa:
Sorry, seems that the above patch does not work on Linux actually. Converting this to draft for now.

yukawa:
Decided to abandon this approach. Let's see how the following one works.

 * https://github.com/google/mozc/pull/1228

PR #1230, opened 2025-03-19
Remove the dependency on PIP from `update_deps.py`

### LINKED ISSUE #1229: Remove the dependency on `venv` from `src/build_tools/update_deps.py`
## Description
Currently the build steps for Android and macOS are somewhat complicated in terms of the fact that we had to explain how to set up `venv` then enable it.

https://github.com/google/mozc/blob/96a8a29fb85a3bf433e08376e3546025857e931a/docs/build_mozc_for_android.md?plain=1#L10-L22

The reason why we need to set up `venv` is because `src/build_tools/update_deps.py` currently uses a PIP module `requests` to download archives. We should be able to remove this dependency by using a standard libraries [`urllib`](https://docs.python.org/3/library/urllib.html) instead.
https://github.com/google/mozc/blob/96a8a29fb85a3bf433e08376e3546025857e931a/src/build_tools/update_deps.py#L181-L199

Let's update `src/build_tools/update_deps.py` so that it can run only with standard libraries.

## Steps to reproduce
1. `git clone https://github.com/google/mozc.git`
2. `cd mozc/src`
3. `python3 build_tools/update_deps.py`

## Expected behavior
The step 3 succeeds, even without `python3 -m pip install requests`

## Version or commit-id
96a8a29fb85a3bf433e08376e3546025857e931a

## Environment
 - OS: All

yukawa:
## Description
With this commit src/build_tools/update_deps.py can run only with Python standard modules without relying on PIP modules. This allows us to simplify our build instructions as we no longer need to explain how to set up python `venv`.

Closes #1229.

## Issue IDs

 * https://github.com/google/mozc/issues/1229

 - OS: All
 - Steps:
   1. Confirm the build is still passing on the GitHub Actions.


jasnliu:
Nice work Yohei, this is Xu, just sent you a request on linkedin, please delete this comment. thx

PR #1245, opened 2025-04-01
Specify `target_compatible_with` to `qt_build_test`

yukawa:
## Description
As a preparation to make unit tests run with Bazel on Windows (#1223), this commit sets `target_compatible_with` to `qt_build_test` so that the test can be excluded when the target is Android, macOS and Windows.

## Issue IDs

 * https://github.com/google/mozc/issues/1223

 - OS: Windows 11 24H2
 - Steps:
   1. `bazelisk build //renderer/qt/... --config oss_windows --build_tests_only`

### DIFF
--- a/src/renderer/qt/BUILD.bazel
+++ b/src/renderer/qt/BUILD.bazel
@@ -145,6 +145,7 @@ mozc_cc_qt_library(
 
 build_test(
     name = "qt_build_test",
+    target_compatible_with = ["@platforms//os:linux"],
     targets = [
         ":qt_ipc_server",
         ":qt_ipc_thread",


src/renderer/qt/BUILD.bazel
```diff
@@ -145,6 +145,7 @@ mozc_cc_qt_library(
 
 build_test(
     name = "qt_build_test",
+    target_compatible_with = ["@platforms//os:linux"],
```
hiroyuki-komatsu, line 148: `os:macos` is also compatible.

It's fine to exclude Windows, but I'm wondering what error happens against this build for Windows.

yukawa: Here is the result. `::setenv` isn't available hence failing. The build succeeds if we drop the line in question though.

```
PS D:\mozc\src> bazelisk --output_base=D:/x test //renderer/qt:qt_build_test --config oss_windows -c dbg --enable_runfiles
INFO: Analyzed target //renderer/qt:qt_build_test (54 packages loaded, 6962 targets configured).
ERROR: D:/mozc/src/renderer/qt/BUILD.bazel:82:19: Compiling renderer/qt/qt_server.cc failed: (Exit 1): clang-cl.exe failed: error executing CppCompile command (from target //renderer/qt:qt_server) D:\mozc\src\third_party\llvm\clang+llvm-20.1.1-x86_64-pc-windows-msvc\bin\clang-cl.exe @bazel-out/x64_windows-dbg/bin/renderer/qt/_objs/qt_server/qt_server.obj.params
renderer/qt/qt_server.cc(109,5): error: no member named 'setenv' in the global namespace
  109 |   ::setenv("QT_QPA_PLATFORM", "xcb", 1);
      |   ~~^
1 error generated.
Target //renderer/qt:qt_build_test failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 8.440s, Critical Path: 4.30s
INFO: 118 processes: 828 action cache hit, 99 internal, 19 local.
ERROR: Build did NOT complete successfully
//renderer/qt:qt_build_test                                     FAILED TO BUILD

Executed 0 out of 1 test: 1 fails to build.
```

> `os:macos` is also compatible.

Do you really want to keep maintaining this build for macOS even though it's never used for production? IIUC, we can stop running `build_qt.py` before running `bazelisk test` on macOS by excluding this from test targets.

yukawa: > Here is the result. `::setenv` isn't available hence failing. The build succeeds if we drop the line in question though.

Another negative side-effect of trying to support this scenario for Windows is that we also need to build Qt for debug builds, which increases the build cycle time, unless we switch to release build when running unit tests only on Windows.

hiroyuki-komatsu: If we can stop running `build_qt.py` for `bazelisk test`, it worth excluding this test from macOS.

Also thank you for sharing the build log on Windows. If removing that line works for Windows, it'd be nice to fix it in another commit.

You don't need to modify this PR, I will merge your PR as-is (i.e. support `os:linux` only)

hiroyuki-komatsu:
We have merged your PR.

PR #1255, opened 2025-04-08
Make `usage_item_array.data` stable

### LINKED ISSUE #1254: `usage_item_array.data` is not stable due to the usage of `std::sort`
## Description
The content of `usage_item_array.data`, which is to be embedded into `mozc.data` and eventually `mozc_server`, is not stable as the data is sorted with `std::sort` instead of `std::stable_sort` in `src/rewriter/gen_usage_rewriter_dictionary_main.cc`.

https://github.com/google/mozc/blob/ecc4e4f92117f9dd8719fc1b57919d2b534f8b10/src/rewriter/gen_usage_rewriter_dictionary_main.cc#L272-L278

## Steps to reproduce
1. `bazelisk build //data_manager/oss:mozc_dataset_for_oss --config oss_windows`
2. `Get-FileHash bazel-bin/data_manager/oss/usage_item_array.data SHA256`

## Expected behavior
The file content of `bazel-bin/data_manager/oss/usage_item_array.data` is the same across platforms (Linux, macOS, and Windows).

## Actual behavior
The file content of `bazel-bin/data_manager/oss/usage_item_array.data` differs  across platforms (Linux, macOS, and Windows).

## Version or commit-id
ecc4e4f92117f9dd8719fc1b57919d2b534f8b10

## Environment
 - OS: All

yukawa:
## Description
Previously the content of `usage_item_array.data` was processed with `std::sort` rather than `std::stable_sort`. As a result, the content can easily vary depending on platforms (to be precise STL implementations).

By switching to `std::stable_sort`, we can assume the same content will be generated across platforms, which would be a huge win when diagnosing conversion related issues.

Note that `gen_usage_rewriter_dictionary_main.cc` runs as a build step. Thus there should be no major performance implications at runtime.

Closes #1254.

## Issue IDs

 * https://github.com/google/mozc/issues/1254

 - OS: All
 - Steps:
   1. `bazelisk //data_manager/oss:mozc_dataset_for_oss --config oss_windows` (on Windows)
   2. `bazelisk //data_manager/oss:mozc_dataset_for_oss --config oss_macos` (on macOS)
   3. `bazelisk //data_manager/oss:mozc_dataset_for_oss --config oss_linux` (on Linux)
   4. Confirm `bazel-bin/data_manager/oss/usage_item_array.data` is the same.

### DIFF
--- a/src/rewriter/gen_usage_rewriter_dictionary_main.cc
+++ b/src/rewriter/gen_usage_rewriter_dictionary_main.cc
@@ -275,7 +275,8 @@ void Convert() {
   LoadUsage(absl::GetFlag(FLAGS_usage_data_file), &usage_entries,
             &conjugation_list);
   RemoveBaseformConjugationSuffix(baseform_map, &usage_entries);
-  std::sort(usage_entries.begin(), usage_entries.end(), UsageItemKeynameCmp);
+  std::stable_sort(
+      usage_entries.begin(), usage_entries.end(), UsageItemKeynameCmp);
 
   // Assign unique index to every string data.  The same string share the same
   // index, so the data is slightly compressed.


hiroyuki-komatsu:
We have merged your PR.

PR #1256, opened 2025-04-09
Normalize `submodules` param in `actions/checkout`

yukawa:
## Description
Removed unnecessary quotes around the `submodules` parameter in the [`actions/checkout`](https://github.com/actions/checkout) step for better readability and consistency.

Note that there are no functional changes, as the quotes were not needed in the first place. This change is purely cosmetic and aims to improve the readability of the workflow files.

## Issue IDs
N/A

 - OS: All
 - Steps:
   1. Confirm all the GitHub Actions are still passing

### DIFF
--- a/.github/workflows/android.yaml
+++ b/.github/workflows/android.yaml
@@ -21,7 +21,7 @@ jobs:
       - name: checkout
         uses: actions/checkout@v4
         with:
-          submodules: 'recursive'
+          submodules: recursive
 
       - name: setup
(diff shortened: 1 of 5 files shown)


hiroyuki-komatsu:
Thank you for your PR. We are going to merge the PR.

By the way, I'm thinking of removal of git submodules. We can use Bazel Modules and update_deps.py instead.

What do you think?

PR #1262, opened 2025-04-13
Stop deploying `mozc_tip64.dll.pdb` again

### LINKED ISSUE #1261: Stop deploying debug symbol for Mozc TIP DLL
## Description
We have been deploying debug symbol for Mozc TIP DLL since October 2024.

 * https://github.com/google/mozc/issues/1081

Now I would propose to revert the above change at least until our migration to Bazel becomes complete.

The main reason to revert back to the original behavior is that `/PDBSTRIPPED` option, which we have been relying on since 94fcce2ecd1058f7028b9c00071e5a20349db74c, has not been supported in `clang-cl` yet. While there most likely can be a way to achieve a similar size reduction, the migration to Bazel should have a higher priority than deploying this optional debug symbol file. Let's revert back to the previous behavior for now.

Another reason to stop deploying the debug symbol is to simplify the steps to support ARM64 in Windows.

## Steps to reproduce
1. Build `Mozc64.msi` with Bazel and install it.
2. `dir %ProgramFiles(x86)%\Mozc\`

## Expected behavior
 * `mozc_tip64.dll.pdb` is not installed.

## Actual behavior
 * `mozc_tip64.dll.pdb` is installed, which is working as expected in the context of issue 1081.

## Version or commit-id
42bbba5777d1c44784f7b4aa734159a6ee80b95d

## Environment
 - OS: Windows 11 24H2

yukawa:
## Description
This commit effectively reverts my previous commit (0377ddd0b617c824b93d17e87bab028acc0ef3e7), which aimed to deploy the PDB file for `mozc_tip64.dll` for better debuggability.

It turns out that `clang-cl` does not support the `/PDBSTRIP` option, which means that a further change is needed to keep the installer size as similar as possible when switching from GYP to Bazel.

Let's undo this change so as not to block the Bazel migration.

Closes #1261

## Issue IDs

 * https://github.com/google/mozc/issues/1261

 - OS: Windows 11 24H2
 - Steps:
   1. Build `Mozc64.msi` then install it.
   2. Confirm `%ProgramFiles(x86)%\Mozc\mozc_tip64.dll.pdb` does not exist.


hiroyuki-komatsu:
The code looks good to me. Thank you for the PR.

IIUC, pdb files can be stripped by using `pbdcopy` with `-p` option.

https://learn.microsoft.com/windows-hardware/drivers/debugger/using-pdbcopy

Since it requires an extra toolchain, committing this PR makes sense rather than introducing `pdbcopy`.
On the other hand, I wondered if this could be an option for the next step.

hiroyuki-komatsu:
Those steps make sense to me too.

We have merged your PR.

Issue #1263, opened 2025-04-14
Remove `mozc::CrashReportHandler` and the dependency on `breakpad`


yukawa:
## Description
This is a preparation to fully migrate to Bazel in macOS and Windows builds.

Currently `mozc::CrashReportHandler` to catch runtime crash to generate minidump with [`breakpad`](https://chromium.googlesource.com/breakpad/breakpad) is supported only in the GYP build with `branding=GoogleJapaneseInput`. This is one of non-trivial behavior differences between the Bazel build and GYP build right now.

To make both builds consistent, basically there are two choices.

 * A. Enable `mozc::CrashReportHandler` for Bazel build with `branding=GoogleJapaneseInput`
 * B. Remove `mozc::CrashReportHandler` from both GYP and Bazel build even with `branding=GoogleJapaneseInput`

What I would propose is to go for B, because

 * Minidump analysis is not actively done these days actually.
 * Chromium has already switched to [`crashpad`](https://chromium.googlesource.com/crashpad/crashpad) from `breakpad`. There is a long standing TODO bug to see if Mozc can also switch to `crashpad`, which has never been realized. 
 * By removing `mozc::CrashReportHandler`, we can of course remove the dependency on `breakpad`.

## Expected behavior
 * `src/base/crash_report_handler*` are removed.
 * `src/third_party/breakpad` is removed.
 * `src/data/installer/credits_en.html` no longer mentions about `google-breakpad`.

## Version or commit-id
dd45cbd6859da868e8aa0c1c6c2b86d74309148f

## Environment
 - OS: Windows and macOS

## Additional context
There must be no observable behavior change in `branding=Mozc` anyway, as we have never enabled crash reporting in the OSS builds.

yukawa:
Done with 1078e63881a86be21114d4faca54dda6353ea439.

PR #1264, opened 2025-04-14
Stop using `CrashReportHandler` under `win32/tip`

yukawa:
## Description
As the first step towards removing `CrashReportHandler` from the codebase (#1263), this commit removes its usage from the `src/win32/tip` directory.

Other directories will be handled in subsequent commits.

Note that there must be no observable behavior change as `CrashReportHandler` has been no-op in the OSS build. It is used when and only when 1) `branding` is `GoogleJapaneseInput` and 2) it is built with GYP.

## Issue IDs

 - OS: Windows 11 24H2
 - Steps:
   1. `bazelisk test ... --config oss_windows -c dbg --build_tests_only`

### DIFF
--- a/src/win32/tip/BUILD.bazel
+++ b/src/win32/tip/BUILD.bazel
@@ -55,11 +55,9 @@ mozc_cc_library(
         ":tip_text_service_impl",
         ":tip_ui_handler",
         ":tip_ui_handler_impl",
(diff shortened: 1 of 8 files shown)


hiroyuki-komatsu:
We have merged your PR.

PR #1282, opened 2025-05-08
Make embedded preedit TSVs consistent

### LINKED ISSUE #1281: Some files under `src/data/preedit/` are embedded only in Bazel build
## Description

The following files are embedded when and only when executables are built with Bazel. These files are not embedded when built with GYP.

 * `src/data/preedit/50keys-hiragana.tsv`
 * `src/data/preedit/flick-number.tsv`
 * `src/data/preedit/toggle_flick-number.tsv`

https://github.com/google/mozc/blob/81dbd2d402b8a8a7c95cade109de36086373343c/src/base/base.gyp#L369-L408

https://github.com/google/mozc/blob/81dbd2d402b8a8a7c95cade109de36086373343c/src/data/preedit/BUILD.bazel#L30-L56

While this discrepancy is not obvious in desktop as all the above missing files are for mobile, let's also embed these files in GYP build as well to minimize the diff between GYP build and Bazel build before fully migrating to Bazel.

## Steps to reproduce
1. Build `mozc_server.exe` with GYP and Bazel.
2. Check `mozc_server.exe` to see if the above three TSV files are embedded or not.

## Expected behavior
The above three TSV files are embedded in both cases.

## Actual behavior
The above three TSV files are embedded only when `mozc_server.exe` is built with Bazel.

## Version or commit-id
81dbd2d402b8a8a7c95cade109de36086373343c

## Environment
 - OS: All

yukawa:
## Description
With this commit the following three preedit TSVs are embedded not only in the executabled built with Bazel but also executables built with GYP.

 * `src/data/preedit/50keys-hiragana.tsv`
 * `src/data/preedit/flick-number.tsv`
 * `src/data/preedit/toggle_flick-number.tsv`

Note that the above missing TSV files are used only in the mobile environment. Basically there should be no user-observable changes except for the binary diff between Bazel build and GYP build becomes smaller.

Here are the file size diffs in the GYP build.

| File Name  |     Old  |     New  |  Diff |
|--------------|--------|---------|-------|
|`mozc_broker.exe`| 1966592  | 1976320 |  9728|
|`mozc_server.exe`|  19974144 | 19984384 | 10240|
|`mozc_tip32.dll`|    4952576  | 4962304 |  9728|
|`mozc_tip64.dll`|    5143552  | 5153280 |  9728|
|`mozc_tool.exe`|     2954240 |  2963968 | 9728|

Ideally the diff should be observed only in the executable where the embedded file is actually used, but it is a different story that needs to be optimized in the future.

Closes #1281.

## Issue IDs

 * https://github.com/google/mozc/issues/1281

 - OS: Windows 11 24H2
 - Steps:
  1. Build `mozc_server.exe` with GYP and Bazel.
  2. Confirm `mozc_server.exe` to see if the above three TSV files are embedded.


hiroyuki-komatsu:
We have merged your PR.

PR #1285, opened 2025-05-12
Stop calling `CrashReportHandler::Initialize`

yukawa:
## Description
To follow up my previous commit (c06d5fcf996c1b9c909f7612fbaa7180e485a6e9), with this commit we stop calling `CrashReportHandler::Initialize` in all the remaining executables.

This is a major milestone towards removing `CrashReportHandler` and the dependency on `breakpad` (#1263).

Note that there must be no observable behavior change in the OSS build, where `CrashReportHandler` has been no-op. It has been funtional when and only when 1) `branding` is `GoogleJapaneseInput` and 2) executables are built with GYP.

`CrashReportHandler` and the dependency on `breakpad` are going to be removed in subsequent commits.

## Issue IDs

 - OS: All
 - Steps:


hiroyuki-komatsu:
We have merged your PR.

PR #1300, opened 2025-05-29
Make `custom_action.cc` buildable for ARM64

yukawa:
## Description
This is a preparation to support Windows ARM64 environment (#1130).

This commit makes it possible for custom_action.cc to be built for ARM64 environment. Actually the DLL path that `TsfRegistrar::RegisterProfiles` takes is a Win32 resource DLL path, which does not need to match the target CPU architecture. So the existing `_M_X64`/`_M_IX86` branchings have been completely unnecessary.

There must be no user observable behavior change.

## Issue IDs

 - OS: Windows 11 24H2
 - Steps:
   2. Confirm that Mozc icon is still shown in the IME picker.
image


src/win32/base/tsf_registrar.h
```diff
@@ -44,7 +46,7 @@ class TsfRegistrar {
 
   // Registers this COM server to the profile store for input processors.
   // The caller is responsible for initializing COM before call this function.
-  static HRESULT RegisterProfiles(const wchar_t *path, DWORD path_length);
+  static HRESULT RegisterProfiles(std::wstring_view icon_path);
```
hiroyuki-komatsu, line 49: Please use the same name for the parameter (icon_path vs resource_dll_path).

src/win32/base/tsf_registrar.cc
```diff
@@ -73,7 +73,7 @@ const GUID kCategories[] = {
 //  5. Click "Details" in the "Text services and input languages" frame, and;
 //  6. All installed prossors are enumerated in the "Installed services"
 //     frame.
-HRESULT TsfRegistrar::RegisterProfiles(const wchar_t *path, DWORD path_length) {
+HRESULT TsfRegistrar::RegisterProfiles(std::wstring_view resource_dll_path) {
```
hiroyuki-komatsu, line 76: Please include the header file for `std::wstring_view`.

hiroyuki-komatsu (commented):
Thank you for your PR.

src/win32/base/tsf_registrar.cc
yukawa: Done.

yukawa:
Rebased and squashed.

PR #1305, opened 2025-06-04
Start building `Mozc64.msi` for ARM64 environment

yukawa:
## Description
This is an important milestone towards supporting ARM64 Windows.

With this commit, we can start building `Mozc64.msi` that is compatible with ARM64 processes and x86 processes on Windows. To build `Mozc64.msi` to be ARM64 compatible, run the following commands:

```
python build_tools/build_qt.py --release --confirm_license --target_arch=arm64
bazelisk build package --config oss_windows --config release_build --platforms=//:windows-arm64
```

Note that no additional Visual Studio components are required unless you explicitly specify the above options. Existing continuous builds for x64 should continue to work without any changes.

Note also that using Mozc in x64 processes on ARM64 machines is not yet supported, as it requires us to build an ARM64X DLL that can be loaded into both x64 and arm64 processes. It will be handled in a subsequent commit.

## Issue IDs

 - OS: Windows 11 24H2 ARM64
 - Steps:
   1. `python build_tools/build_qt.py --release --confirm_license --target_arch=arm64`
   2. `bazelisk build package --config oss_windows --config release_build --platforms=//:windows-arm64`
   3. `python build_tools/open.py bazel-bin/win32/installer/Mozc64.msi`
   4. Confirm that you can use Mozc on Notepad.

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -74,7 +74,7 @@ jobs:
           path: src/out_win/Release/Mozc64.msi
           if-no-files-found: warn
 
-  build:
+  build_x64:
     # https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md
     runs-on: windows-2022
     timeout-minutes: 120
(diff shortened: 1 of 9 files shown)


hiroyuki-komatsu:
We have merged your PR.

PR #1312, opened 2025-06-13
Remove unused `crash_report_handler` dependency

yukawa:
## Description
This follows up to my previous commit (18c59459ad4f5a3aa53b825e122d38eceffd744c), which remove all the remaining invocations of `CrashReportHandler::Initialize` from the codebase towards removing the dependency on `breakpad` (#1263).

One thing I missed in that commit was the `//base:crash_report_handler` dependency in `src/mac/BUILD.bazel` and `src/server/BUILD.bazel` files were not cleaned up. As a result, the `//base:crash_report_handler` target is still being built, even though it is not used anywhere in the codebase. Let's also remove those unused dependencies before we fully remove `src/base/crash_report_handler.*` files.

## Issue IDs

 - OS: All
 - Steps:
   1. All the GitHub Actions pass

### DIFF
--- a/src/mac/BUILD.bazel
+++ b/src/mac/BUILD.bazel
@@ -125,7 +125,6 @@ mozc_objc_library(
         ":mozc_imk_input_controller",
         ":renderer_receiver",
         "//base:const",
-        "//base:crash_report_handler",
         "//base:init_mozc",
         "//base:run_level",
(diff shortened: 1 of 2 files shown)


hiroyuki-komatsu:
We have merged your PRs.

Issue #1318, opened 2025-07-01
Using  `_SC_GETPW_R_SIZE_MAX` cause crash in some Linux systems


mocknen:
Not every Linux systems have `_SC_GETPW_R_SIZE_MAX`. In such an environment, the following code will cause abortions.

https://github.com/google/mozc/blob/22ccfeb9fe04ea2b72da0d455e5a5423b0e8554a/src/base/system_util.cc#L529-L530

For reference, the following code is a golang handling of this issue.

https://github.com/golang/go/blob/acb914f2c2a3ec3dd227812ba6b119480d6400cd/src/os/user/cgo_lookup_unix.go#L151-L157

yukawa:
Closing as #1318 got merged.

PR #1319, opened 2025-07-01
Use fallback value when `_SC_GETPW_R_SIZE_MAX` is not defined

mocknen:
## Description

`_SC_GETPW_R_SIZE_MAX` is not always defined in Linux systems. If sysconf returns -1, fallback to default value.

## Issue IDs

#1318


hiroyuki-komatsu:
We have merged your PR.

Issue #1295, opened 2025-05-28
Simplify build config transitions in Windows Bazel build


yukawa:
## Description

Currently `mozc_win32_cc_prod_binary` defines a local platform every time it is used.
https://github.com/google/mozc/blob/e155855b4b840f68d790c7572f4fed0f7a730fc8/src/build_defs.bzl#L427-L435

This is turned out to be quite inefficient as it looks like Bazel does not automatically treat two ore platforms that have the same `constraint_values` as the same platform. As a result, the following targets build their dependencies separately, which is unnecessarily slowing down our build.
 * `mozc_server.exe`
 * `mozc_renderer.exe`
 * `mozc_cache_service.exe`
 * `mozc_broker.exe`
 * `mozc_tool.exe`
 * `mozc_installer_helper.dll`

It should be much more efficient if we globally define platforms at the top level and use them when having to specify the build target.
```
platform(
    name = "windows-x86_32",
    constraint_values = [
        "@platforms//cpu:x86_32",
        "@platforms//os:windows",
    ],
)

platform(
    name = "windows-x86_64",
    constraint_values = [
        "@platforms//cpu:x86_64",
        "@platforms//os:windows",
    ],
)
```

Another potential issue is that how we are going to support building two (or more) installers when supporting both x64 and ARM64.

 * https://github.com/google/mozc/issues/1130

To natively support ARM64 environment, we would like to build the above executables (e.g. `mozc_server.exe`) to be ARM64 rather than x64. This is currently cumbersome as we explicitly specify the target CPU architecture for every production binary as follows.
https://github.com/google/mozc/blob/e155855b4b840f68d790c7572f4fed0f7a730fc8/src/server/BUILD.bazel#L63-L69

To simplify the build instructions, a better approach would be to rely on the standard `--platforms` command line options to specify the target architecture.

```
# to build x64 installer
python build_tools/build_qt.py --release --confirm_license --target_arch=x64
bazelisk build --config oss_windows --config release_build package --platforms=//:windows-x86_64
```
```
# to build ARM64 installer
python build_tools/build_qt.py --release --confirm_license --target_arch=arm64
bazelisk build --config oss_windows --config release_build package --platforms=//:windows-arm64
```

This will make the build process much similar to macOS build, where we can specify the target architecture with `--macos_cpus` option.

To summarize, let's
  
 * define `platform` globally then refer to it from `mozc_win32_cc_prod_binary` rather than defining a local private `platform` for every executable.
[...]

PR #1323, opened 2025-07-01
Inline `//:intel_cpu` for simplicity

yukawa:
## Description
This follows up to my previous commit (d5bb10c474ff206f1dde1551ec5ae4cc91677e09), which introduced a synthesized config `//:intel_cpu` that is set when either `@platforms//cpu:x86_32` or `@platforms//cpu:x86_64` is set as part of the effort to speed up our Bazel-based build for Windows (#1295).

The problem is that in downstream targets we can still write select rules without `//:intel_cpu` as follows:

```starlark
select({
  "@platforms//cpu:x86_32": ["/CETCOMPAT"],
  "@platforms//cpu:x86_64": ["/CETCOMPAT"],
  "//conditions:default": [],
}),
```

Let's just write it be written as a redundant but more explicit select rules for now.

## Issue IDs

 * https://github.com/google/mozc/issues/1295

 - OS: Windows 11 24H2
 - Steps:
   1. Confirm you can still build, install, and use Mozc.

### DIFF
--- a/src/BUILD.bazel
+++ b/src/BUILD.bazel
@@ -131,14 +131,6 @@ platform(
     ],
 )
 
-selects.config_setting_group(
-    name = "intel_cpu",
-    match_any = [
-        "@platforms//cpu:x86_32",
-        "@platforms//cpu:x86_64",
-    ],
-)
-
 # Special target so as to define special macros for each platforms.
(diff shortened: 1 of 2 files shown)


hiroyuki-komatsu:
We have merged your PR.

PR #1328, opened 2025-07-14
bzlmod: Use `http_archive` over `archive_override`

yukawa:
## Description
This aims to simplify 'MODULE.bazel' without changing anything.

As explained in the reference, [`archive_override`](https://bazel.build/docs/bzlmod#archive_override) is implemented on top of `http_archive`, and they are actually interchangeable. The issue is that currently `MODULE.bazel` uses both approaches, which is confusing because future readers may wonder which one should be used and there might be a reason for using both.

To avoid such a confusion, this commit replaces `archive_override` with `http_archive` in `MODULE.bazel` for now.

## Issue IDs
N/A

 - Steps:
   1. GitHub Actions still pass

### DIFF
--- a/src/MODULE.bazel
+++ b/src/MODULE.bazel
@@ -169,23 +169,6 @@ new_local_repository = use_repo_rule(
     "new_local_repository",
 )
 
-# Japanese Usage Dictionary (2025-01-25)
-bazel_dep(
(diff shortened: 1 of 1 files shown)


hiroyuki-komatsu:
We have merged your PR.

Issue #1320, opened 2025-07-01
Stop using git submodules


yukawa:
## Description
Managing dependencies on external C++ projects has always been a challenge.

Back in 2010 OSS Mozc originally started with borrowing the [`gclient` command](https://chromium.googlesource.com/chromium/tools/depot_tools.git/+/HEAD/README.gclient.md) and `DEPS` file mechanism from the Chromium project.

In 2015, we switched to git-submodules to remove the dependency on `gclient` command (#300).

In 2023 we also introduced a wrapper script `src/build_tools/update_deps.py` (6efc6490d0ef9b3a0aef95b356ba23481289be33), which then started being used to utilize GitHub Actions Runner cache (8233b4067f5df89026e13ff771ebbc2fa59fb95f).

In 2024, we also started using **bzlmod** in Bazel (#1002), which basically offers all the features we discussed above (downloading archives then extracting them out, importing libraries from module repository, configuring cache directory).

So it's time to start sorting out duplicated functionality there.

The easiest part is git-submodules. For instance, we can already build Mozc with Bazel without the following git submodules as these dependencies are now described in `src/MODULE.bazel`.

 * `src/third_party/abseil-cpp`
 * `src/third_party/protobuf`
 * `src/third_party/gtest`
 * `src/third_party/japanese_usage_dictionary`

As for the remaining submodule-based dependencies, it should not be that difficult for us to fully stop using as long as we use Bazel to build Mozc. As for remaining special use cases in GYP-based builds, we can still take care of them in `src/build_tools/update_deps.py` with some additional option as a short-term solution until we fully stop supporting GYP-based builds.

## Expected behavior
Git submodules are not used in Mozc git repository.

## Actual behavior
Git submodules are still used in Mozc git repository.

## Version or commit-id
22ccfeb9fe04ea2b72da0d455e5a5423b0e8554a

yukawa:
Done with 32be06076656f135588d749c9a8f32123bed8de5 and efaae552cbb5a20f5bf05c8554d8a18b1bcf9063.

PR #1367, opened 2025-09-17
Remove remaining references to breakpad

### LINKED ISSUE #1263: Remove `mozc::CrashReportHandler` and the dependency on `breakpad`

yukawa:
## Description
With the removal of `mozc::CrashReportHandler` (b7f08d61b838ca34234c790b9f6097804074d306), `breakpad` is no longer used even when `branding=GoogleJapaneseInput` in the GYP build. It should now be safe to remove all the remaining references to `breakpad` from the codebase.

Removing `breakpad` is also a step toward removing the dependency on git submodules (#1320), as it has never been used in the bazel build.

Closes #1263.

## Issue IDs

 - OS: All
 - Steps:


hiroyuki-komatsu:
Thank you for your PR. Since I also already have a similar change and this removal requires some internal changes, I will apply my change first. However, I probably need several days to make it happen.

At this moment, please keep this PR as-is.

Thank you for your cooperation.

yukawa:
Closing as this is already done by 1078e63881a86be21114d4faca54dda6353ea439.

Issue #1376, opened 2025-09-29
Consider deprecating "restricted mode" from `mozc_server` and `mozc_renderer` on Windows


yukawa:
## Description

As Mozc no longer supports Windows 7 and prior OSes where nested Win32 Jobs were not supported, chances are quite high that we can remove so-called **Restricted Mode** from `mozc_server` and `mozc_renderer`. Let's see if we can actually do that.

https://github.com/google/mozc/blob/2efeece0ae519a17ffceaa766b5be1525f7cec7b/src/server/mozc_server.cc#L52
https://github.com/google/mozc/blob/2efeece0ae519a17ffceaa766b5be1525f7cec7b/src/renderer/renderer_server.cc#L61-L65

### Background

When Mozc's sandbox architecture was originally designed for Windows, one of the challenges was that [nested Win32 Jobs](https://learn.microsoft.com/en-us/windows/win32/procthread/nested-jobs) were not supported, while Mozc wanted to create a Job when launching `mozc_server` and `mozc_renderer` so that those two processes could run in limited resource accesses. The solution we chose was to introduce a special mode called "Restricted Mode" where those processes run without Win32 Job sandboxing only for a limited time duration (e.g. 60 sec). Restricted mode is used only when all the following conditions are met:

 * The server process (`mozc_server` or `mozc_renderer`) is not yet running.
 * The client process (e.g. `notepad.exe`) is already running in a Win32 Job without `JOB_OBJECT_LIMIT_BREAKAWAY_OK` nor `JOB_OBJECT_LIMIT_SILENT_BREAKAWAY_OK`.

https://github.com/google/mozc/blob/2efeece0ae519a17ffceaa766b5be1525f7cec7b/src/client/server_launcher.cc#L118-L134

Now that nested Win32 Jobs have been fully supported since Windows 8, it's highly likely the above special mode is no longer necessary.

## Steps to reproduce
TBD.

## Expected behavior
TBD.

## Actual behavior
TBD.

## Version or commit-id
2efeece0ae519a17ffceaa766b5be1525f7cec7b

## Environment
 - OS: Windows 24H2

PR #1368, opened 2025-09-19
Remove unused `set_restricted()` method

yukawa:
## Description
As a preparation to remove the "restricted mode" (#1376), let's start from removing the following unused methods.

 - `ServerLauncherInterface::set_restricted()`
 - `ClientInterface::set_restricted()`

## Issue IDs

 * https://github.com/google/mozc/issues/1376

## Steps to test new behaviors
 - OS: All
 - Steps:
   1. Confirm all GitHub Actions pass

PR #1379, opened 2025-09-30
Always let server/renderer run in Win32 Jobs

yukawa:
## Description
As the initial step towards removing "restricted mode" (#1376), this commit simplifies the codebase by taking advantage of the fact that Windows 8 and later allows Win32 processes to belong to multiple jobs.

Before Windows 8, Win32 processes cannot belong to multiple jobs. Due to this limitation, there is a case when `mozc_{server,renderer}` cannot be launched in a restricted job. This is why `WinSandboxSecurityInfo::use_locked_down_job` has been determined at runtime with `RunLevel::IsProcessInJob()`.

Now that Mozc only supports Windows 10 and later, `use_locked_down_job` can be initialized as the compile time constant.

To minimize the compatibility impact, `CREATE_BREAKAWAY_FROM_JOB` flag continues to be specified if possible. This means that when Mozc server/renderer is launched from a process in a job which allows breakaway, the user observable behavior remains unchanged before and after this commit.

## Issue IDs

 * https://github.com/google/mozc/issues/1376

 - OS: Windows 11 24H2
 - Steps:
   1. Build and install Mozc
   2. Open Notepad
   3. Confirm you can still enable Mozc and use it.

### DIFF
--- a/src/base/run_level.cc
+++ b/src/base/run_level.cc
@@ -276,32 +276,6 @@ RunLevel::RunLevelType RunLevel::GetRunLevel(RunLevel::RequestType type) {
 #endif  // _WIN32
 }
 
-bool RunLevel::IsProcessInJob() {
-#ifdef _WIN32
-  // Check to see if we're in a job where
-  // we can't create a child in our sandbox
-
-  JOBOBJECT_EXTENDED_LIMIT_INFORMATION JobExtLimitInfo;
-  // Get the job information of the current process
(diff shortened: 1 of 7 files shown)


yukawa:
Converting this to draft to make it clear that #1368 needs to be merged first to avoid merge conflict.

yukawa:
Rebased on top of ff6a842c096cf7f8e8b3903d666f966106b9728f. Now this is ready to be reviewed.

Issue #1296, opened 2025-05-28
Support building Mozc on Windows ARM64 environment.


yukawa:
## Description
This is a stretch goal of supporting Windows ARM64 environment.

## Expected behavior
* We can build Mozc on Windows ARM64 environment.
* There is a continuous build to ensure Mozc can be built on Windows ARM64 environment.

## Actual behavior
* Building Mozc on Windows ARM64 environment is not supported.

## Version or commit-id
e155855b4b840f68d790c7572f4fed0f7a730fc8

## Environment
 - OS: Windows 24H2 ARM64

PR #1382, opened 2025-10-08
Use ARM64-native LLVM binaries on Windows ARM64

yukawa:
## Description
Previously, `update_deps.py` always downloaded x86_64 LLVM toolchains for Windows, even on ARM64 machines. While these binaries can still run under WoA emulation, it is preferable to use ARM64-native binaries when possible for better performance. Let's do so as part of the ongoing efforts to support ARM64 Windows machines.

With this commit, `update_deps.py` starts downloading platform-native LLVM toolchains rather than always downloading x86_64 ones.

Note that LLVM version remains unchanged to be 20.1.1. Thus there should be no bit-level differences for the generated binaries.

## Issue IDs

 - OS: Windows 11 24H2 (ARM64)
 - Steps:
   1. `python3 build_tools/update_deps.py`
   2. Confirm `third_party/llvm/bin/clang-cl.exe` is an ARM64 native binary.

### DIFF
--- a/src/build_tools/update_deps.py
+++ b/src/build_tools/update_deps.py
@@ -40,6 +40,7 @@ import dataclasses
 import hashlib
 import os
 import pathlib
+import platform
 import shutil
 import stat
 import subprocess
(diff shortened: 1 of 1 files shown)


yukawa:
Converting this to a draft as 30bb1588c2011ebbfc7ba50840cdf43a47de7905 got reverted as 1d5e28d3eb8ce9b0ac5436310c14449968205b68.

This commit depends on 30bb1588c2011ebbfc7ba50840cdf43a47de7905.

yukawa:
@hiroyuki-komatsu  With 7f10d8341aa61a43bb9f575fd0c974843153a68a, this PR is now ready for review. Thanks!

hiroyuki-komatsu:
We have merged your PR.

PR #1385, opened 2025-10-09
Fix text reconversions on Excel for Windows

### LINKED ISSUE #1384: Reconversion no longer works on Excel for Windows
## Description
The reconversion command (e.g. <kbd>変換</kbd>) no longer works on Excel for Windows.
This is a recent regression in a2af4322f5863e14495291fdcdf5b9bad62d7e6a while working on the following feature request.

 * #1289

## Steps to reproduce
2. Have an Excel sheet where some characters are already filled in a cell.
3. Focus into the cell with characters without selecting characters.
4. Enable Mozc
5. Hit a key (combination) that is associated with `Reconvert`.

## Expected behavior
Image
* Characters are selected then being reconverted.

## Actual behavior
Image
* Characters are selected but reconversion doesn't happen.

## Version or commit-id
a2af4322f5863e14495291fdcdf5b9bad62d7e6a  and any later versions.

## Environment
 - OS: Windows 11 25H2
 - Related Applications: Excel 2509 (Build 19231.20156)

yukawa:
## Description
This follows up my previous commit (a2af4322f5863e14495291fdcdf5b9bad62d7e6a) , which aimed to support surrounding text retrieval on Chromium-based apps. However, it turned out that it broke text reconversion on Excel. This commit aims to fix the text reconversion on Excel while keeping the support for surrounding text retrieval on Chromium-based apps.

A short takeaway is that Excel has relied on `ITfFnReconversion`, where the application (Excel) is responsible for providing `ITfRange` object from which IMEs can retrieve the surrounding text. Special rules implemented in `TipSurroundingText::Get` are not necessary when a valid `ITfRange` object is already given. This is what I have overlooked even before the culprit commit.

This commit utilizes the existing `TipEditSession::GetTextSync` as it already takes an `ITfRange` object to return the selected text. For now reusing it would be the simplest solution.

Closes #1384.

## Issue IDs

 * https://github.com/google/mozc/issues/1384

 - OS: Windows 11 25H2,  Excel 2509
 - Steps:
reconvert_excel_expected

### DIFF
--- a/src/win32/tip/tip_edit_session.cc
+++ b/src/win32/tip/tip_edit_session.cc
@@ -49,6 +49,7 @@
 #include "win32/base/conversion_mode_util.h"
 #include "win32/base/deleter.h"
 #include "win32/base/input_state.h"
+#include "win32/tip/tip_composition_util.h"
 #include "win32/tip/tip_dll_module.h"
 #include "win32/tip/tip_edit_session_impl.h"
 #include "win32/tip/tip_input_mode_manager.h"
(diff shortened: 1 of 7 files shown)


src/win32/tip/tip_edit_session.cc
```diff
@@ -553,23 +554,42 @@ class SyncGetTextEditSessionImpl final
...
   // request is granted.
   virtual STDMETHODIMP DoEditSession(TfEditCookie read_cookie) {
-    TipRangeUtil::GetText(range_.get(), read_cookie, &text_);
+    HRESULT result = TipRangeUtil::GetText(range_.get(), read_cookie, &text_);
+    if (FAILED(result)) {
+      return result;
+    }
+    if (check_composition_) {
+      wil::com_ptr_nothrow<ITfContext> context;
+      result = range_->GetContext(&context);
```
hiroyuki-komatsu: Is `GetContext` lightweight? If no, how about propagating the context from the caller (i.e. `GetTextSync`)?

src/win32/tip/tip_edit_session.cc
```diff
@@ -553,23 +554,42 @@ class SyncGetTextEditSessionImpl final
...
+      }
+      wil::com_ptr_nothrow<ITfCompositionView> composition_view =
+          TipCompositionUtil::GetCompositionView(context.get(), read_cookie);
+      is_composing_ = !!composition_view;
+    }
     return S_OK;
   }
 
   const std::wstring &text() const { return text_; }
+  const bool is_composing() const { return is_composing_; }
```
hiroyuki-komatsu: Please use `bool is_composing` (without `const`).

yukawa: Oops. Fixed.

hiroyuki-komatsu:
We have merged your PR.

Issue #1424, opened 2026-01-10
Emacs: Candidate window not displayed due to protocol incompatibility between mozc.el and mozc_emacs_helper


kiyoka:
## Description
Conversion candidates are not displayed in the candidate window when using mozc in Emacs, even though the conversion functionality itself works correctly. This issue occurs due to protocol incompatibility between newer versions of mozc.el and older versions of mozc_emacs_helper.

## Steps to reproduce
1. Install mozc_emacs_helper v2.29.5160.102 or earlier via system package manager (e.g., apt, dnf, pacman)
2. Install or update mozc.el v2.30.5618 or later from MELPA
3. Open Emacs and start Japanese input with mozc
4. Type some text and press Space to convert (e.g., type "henkan" and press Space)
5. Observe that conversion occurs but the candidate window does not appear

## Expected behavior
When converting Japanese text, the candidate window should appear showing multiple conversion candidates that the user can select from.

## Actual behavior
The conversion happens (text changes), but the candidate window remains invisible. Users cannot see or select alternative conversion candidates.

## Screenshots
N/A (See detailed blog post for more information: https://zenn.dev/kiyoka/articles/emacs-mozc-version-upgrade-issue)

## Version or commit-id
- Affected older version: mozc_emacs_helper v2.29.5160.102 or earlier
- Affected newer version: mozc.el v2.30.5618 or later (from MELPA)
- Breaking change introduced in commit: [ec4ff27eb](https://github.com/google/mozc/commit/ec4ff27eb) (October 21, 2024)

## Environment
- OS: Linux (affects various distributions including Ubuntu, Fedora, Arch Linux, etc.)
- IMF (for Linux): N/A (This issue is specific to Emacs integration)
- Related Applications: GNU Emacs with mozc.el

## Investigations
- Whether this issue happens on both Ibus and other IMF (e.g. Fcitx, uim).
  + N/A (This issue is specific to the Emacs frontend, not IMF-related)
- Whether this issue happens on other IMEs (e.g. Anthy, SKK).
  + N/A (Issue is specific to mozc protocol change)
- What applications this issue happens on (e.g. Chromium, gedit).
  + GNU Emacs with mozc.el package
- What applications this issue does not happen on (e.g. Chromium, gedit).
  + Other applications using Ibus/Fcitx frontends (not affected)
- (optional) What versions or commit-ids this issue did not happen on
  (e.g. Mozc-2.28.4960.100+24.11.oss).
[...]

PR #1427, opened 2026-01-12
Fix backward compatibility for candidate window display in mozc.el

kiyoka:
## Description

This PR fixes a backward compatibility issue where mozc.el cannot display the candidate window when used with older versions of mozc_emacs_helper.

After investigating, I confirmed with the Mozc developers that the protocol change in commit ec4ff27eb was simply a field rename from `candidates` to `candidate-window` with no functional changes.

The fix modifies mozc.el to support both field names, ensuring it works with both old and new versions of mozc_emacs_helper without breaking existing setups.

## Issue IDs

#1424

- OS: [e.g. Linux with Emacs]
- Steps:
  1. Use mozc.el (v2.30.5618 or later) with an older mozc_emacs_helper (v2.29.5160.102 or earlier)
  2. Activate mozc-mode and start typing in Japanese
  3. Confirm that the candidate window appears correctly
  4. Check the *Messages* buffer for debug output showing which field is being used

## Additional context

This change ensures backward compatibility by checking for `candidate-window` first (new protocol) and falling back to `candidates` (old protocol) if not present. Debug logging has been added to help identify which protocol version is being used.

### DIFF
--- a/src/unix/emacs/mozc.el
+++ b/src/unix/emacs/mozc.el
@@ -343,13 +343,13 @@ EVENT is the last input event, which is usually passed by the command loop."
 
        ;; Mozc server consumed the key event.
        ((mozc-protobuf-get output 'consumed)
-        (let* ((result (mozc-protobuf-get output 'result))
-               (preedit (mozc-protobuf-get output 'preedit))
-               ;; Support both 'candidate-window (new) and 'candidates (old).
-               ;; 'candidates is also checked for backward compatibility with
-               ;; older versions of mozc_emacs_helper (see: #1424).
-               (candidates (or (mozc-protobuf-get output 'candidate-window)
-                               (mozc-protobuf-get output 'candidates))))
+        (let ((result (mozc-protobuf-get output 'result))
+              (preedit (mozc-protobuf-get output 'preedit))
+              ;; Support both 'candidate-window (new) and 'candidates (old).
+              ;; 'candidates is also checked for backward compatibility with
+              ;; older versions of mozc_emacs_helper (see: #1424).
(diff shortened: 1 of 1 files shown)


kiyoka:
I have signed the CLA. Could you please re-run the CLA check?

src/unix/emacs/mozc.el
```diff
@@ -343,9 +343,15 @@ EVENT is the last input event, which is usually passed by the command loop."
...
-        (let ((result (mozc-protobuf-get output 'result))
-              (preedit (mozc-protobuf-get output 'preedit))
-              (candidates (mozc-protobuf-get output 'candidate-window)))
+        (let* ((result (mozc-protobuf-get output 'result))
+               (preedit (mozc-protobuf-get output 'preedit))
+               (cand-window (mozc-protobuf-get output 'candidate-window))
+               (cands (mozc-protobuf-get output 'candidates))
+               (candidates (progn
+                             (when (or cand-window cands)
+                               (message "mozc.el: Using '%s' field for candidates"
```
hiroyuki-komatsu: Does this show a message every key event?

How about simplifying as follows?

```elisp
(candidates (or (mozc-protobuf-get output 'candidate-window)
                (mozc-protobuf-get output 'candidates)))
```

hiroyuki-komatsu (commented):
Thank you for your contribution.

kiyoka:
Updated per your suggestions. Tested with both old and new mozc_emacs_helper versions.

hiroyuki-komatsu:
The PR looks good to me.
I'm working on merging it. Please wait a moment.

Thank you,

hiroyuki-komatsu:
We have merged the PR. Thank you for your contribution!

Issue #1430, opened 2026-01-21
Create a universal installer for both Windows x64 and ARM64


yukawa:
## Description
In #1130, we can already build an ARM64 native installer for Windows. However, it would also be nice if we can have a universal install that supports both Windows x64 and ARM64 machines so that we can reuse the existing deploy pipeline. 

## Expected behavior
We can build a universal installer that supports both Windows x64 and ARM64 machines. Deploying x64 executables (e.g. `mozc_server.exe`) to ARM64 machines is acceptable for the sake of reusing existing deployment channels.

## Version or commit-id
dab80b024f2286590b381b59c1a67821084328b7

## Environment
 - OS: Windows 11 25H2

PR #1432, opened 2026-01-21
Add `defines` to `mozc_win32_cc_prod_binary` input

yukawa:
## Description
As a preparation to support universal installer for Windows x64 and ARM64 (#1430), this commit adds `defines` parameter to `mozc_win32_cc_prod_binary` rule so that we can later define a preprocessor macro only when universal installer is explicitly enabled.

At this point, the new parameter is not used yet. There must be no functional change after this commit.

## Issue IDs

 * https://github.com/google/mozc/issues/1430

 - OS: Windows 11 25H2
 - Steps:
   1. Confirm all the GitHub Actions pass.

### DIFF
--- a/src/build_defs.bzl
+++ b/src/build_defs.bzl
@@ -319,6 +319,7 @@ def mozc_win32_cc_prod_binary(
         name,
         executable_name_map = {},  # @unused
         srcs = [],
+        defines = [],
         deps = [],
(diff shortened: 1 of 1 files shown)


hiroyuki-komatsu:
We have merged your PRs.
Thank you for your contributions!

PR #1446, opened 2026-03-17
fix: handle symlinked `/usr/lib` during server validation

neoto:
## Description
On distros where `/usr/lib` is a symlink to `/usr/lib64` (e.g. Solus), `readlink(/proc/<pid>/exe)` returns the canonicalized path, which won't match the comptime `MOZC_SERVER_DIR`. Using `FileUtil::IsEquivalent` to compare inodes helps in this case.

## Issue IDs
I thought this was a straightforward change so I didn't create an issue, but I can do so if preferred.

There isn't really a test to verify this behavior, nor am I sure what we should be testing for. I'm open to any suggestions!

よろしくお願いします！

### DIFF
--- a/src/ipc/ipc_path_manager.cc
+++ b/src/ipc/ipc_path_manager.cc
@@ -412,6 +412,19 @@ bool IPCPathManager::IsValidServer(uint32_t pid,
   }
 
 #ifdef __linux__
+  // On some Linux distributions (Solus, for example), /usr/lib is a symlink to
+  // /usr/lib64. readlink(/proc/<pid>/exe) always returns the fully
+  // canonicalized path, while the comptime configured server_path may contain
+  // unresolved symlinks.
+  const std::string expected_server_path(server_path);
+  const auto is_equivalent =
+      FileUtil::IsEquivalent(server_path_, expected_server_path);
+
+  if (is_equivalent.value_or(false)) {
+    server_path_ = expected_server_path;
+    return true;
+  }
+
   if (absl::StrCat(server_path, " (deleted)") == server_path_) {
     LOG(WARNING) << server_path << " on disk is modified";
     // If a user updates the server binary on disk during the server is running,


hiroyuki-komatsu:
Hi neoto,
Thank you for your PR!

That is an intentional check to confirm if the server is valid or not.

You can change the value `LINUX_MOZC_SEVER_DIR` in [config.bzl](https://github.com/google/mozc/blob/master/src/config.bzl) to fit your environment.

```
LINUX_MOZC_SERVER_DIR = "/usr/lib/mozc"
```

If it doesn't work, please feel free to open a discussion.

Best,

yukawa:
It seems that the original reporter came here because of a regression https://github.com/getsolus/packages/issues/8202 caused by https://github.com/getsolus/packages/pull/8134.

The root cause of https://github.com/getsolus/packages/issues/8202  is simple. The previous Mozc package (2.29.5150.102) in Solus was configured with explicitly specifying `--server_dir=/usr/lib64/mozc`.
```
    GYP_DEFINES="ibus_mozc_path=/usr/lib64/ibus-mozc/ibus-engine-mozc \
        use_libprotobuf=1 \
        use_libzinnia=1 \
        use_system_abseil_cpp=1" \
    python3 src/build_mozc.py gyp --server_dir=/usr/lib64/mozc --target_platform=Linux
```

The new Mozc package (3.33.6089.102) in Solus, however, misses the corresponding configuration in config.bzl. I suspect it's just something accidentally dropped when the Solus package maintainer migrated from GYP to Bazel.
```
    # Use sources for pre-download files and override registry with local mirror.
    # bazel only respects CC from environment, not CXXFLAGS and LDFLAGS.
    # Pass them with --cxxopt, and --linkopt.
    bazel build \
        --config oss_linux \
        --config stable_channel \
        --config release_build \
        --copt="-DNDEBUG" \
        --copt="-U_FORTIFY_SOURCE "\
        --copt="-Wno-uninitialized" \
        --host_copt="-Wno-uninitialized" \
        --linkopt=-fuse-ld=bfd \
        $(echo "${CFLAGS}"|xargs -n1 echo "--conlyopt") \
        $(echo "${CXXFLAGS}"|xargs -n1 echo "--cxxopt") \
        $(echo "${LDFLAGS}"|xargs -n1 echo "--linkopt") \
        --nostart_end_lib \
        --verbose_failures \
        ${mozc_targets[@]}
```

neoto:
Thank you both for the kind response, and for even doing research. I figured that passing the server directory would fix the issue at hand, but my change ensures that distro maintainers don't really have to think about it at all. I'll ping Evan so that he's aware of this discussion in any case.

Feel free to close this PR if you feel like it doesn't improve the status quo.

hiroyuki-komatsu:
Since Mozc is designed to run from a specific file location, we do not plan to support executing it via symbolic links. We believe this restriction ensures the client communicates with the correct server and minimizes the risk of the server entering an invalid state.

For these reasons, I will be closing this PR. Thank you for your understanding.

However, if you would like to add better logging or error messages to clarify this behavior for users, we would be happy to review a PR for that.

Thank you,

Issue #1437, opened 2026-02-04
Migrate to Bazel 9


yukawa:
## Description
Like we did in the last year to switch from Bazel 7.x to Bazel 8.x, we need some extra work to fully migrate to Bazel 9. 

 * https://github.com/google/mozc/issues/1118

Remaining tasks are:
  
 * Make sure all the dependent Bazel modules are compatible with Bazel 9.x
   * Update module versions as needed.
 * Make sure Mozc's own build rules are compatible with Bazel 9.x
   * Update Mozc's own build rules as needed.
 * Once Mozc becomes fully compatible with Bazel 9.x, drop Bazel 8.x support.
 * (optional) Evaluate the implications of prebuilt proto compiler:
   * https://protobuf.dev/news/2026-01-16/

## Version or commit-id
76cb1ce02ec4cad064c708dd400071f7cb575513

## Additional Context
Right now, the Mozc team is verifying that the Windows and macOS binaries built with Bazel are production-ready, which is the final step before we can deprecate GYP. Since our primary focus is replacing GYP, we want to avoid introducing build tool upgrades that might destabilize the environment during this critical verification phase. Once the GYP deprecation is finalized, we will be in a better position to update dependencies like Bazel.

Thus pull requests that try to update bzlmod modules are unlikely to be merged until we announce that GYP build is fully deprecated.

yukawa:
Should be done with 3ced4c809b43db66a1068b8a544300567668d8d3.

PR #1453, opened 2026-03-31
Make build rules compatible with Bazel 8.6 and 9.0

yukawa:
## Description
As a preparation to Bazel 9 migration, this commit makes the build rules compatible with both Bazel 8.6 and 9.0.

 * Added missing includes for `cc_library` and `objc_library`
 * Removed the dependency on deprecated `local_config_platform`

There must be no change in the final artifacts.

## Issue IDs

 * https://github.com/google/mozc/issues/1437

 - OS: All
 - Steps:


hiroyuki-komatsu:
We have merged your PR.

PR #1460, opened 2026-04-07
Use ARM64-native Ninja binaries on Windows ARM64

yukawa:
## Description
Previously, `update_deps.py` always downloaded x86_64 Ninja executable for Windows, even on ARM64 machines. While these binaries can still run under WoA emulation, it is preferable to use ARM64-native binaries when possible for better performance. Let's do so as part of the ongoing efforts to support ARM64 Windows machines.

With this commit, `update_deps.py` starts downloading platform-native Ninja executable rather than always downloading x86_64 one.

Note that Ninja version remains unchanged to be 1.13.2. Thus there should be no bit-level differences for the generated binaries.

## Issue IDs

 - OS: Windows 11 25H2 ARM64
 - Steps:
   1. `python build_tools/update_deps.py`
   2. `python build_tools/build_qt.py --release --confirm_license`
   3. Confirm `ninja.exe` is running as an ARM64 process.

### DIFF
--- a/src/build_tools/update_deps.py
+++ b/src/build_tools/update_deps.py
@@ -112,6 +112,12 @@ NINJA_WIN = ArchiveInfo(
     sha256='07fc8261b42b20e71d1720b39068c2e14ffcee6396b76fb7a795fb460b78dc65',
 )
 
+NINJA_WIN_ARM64 = ArchiveInfo(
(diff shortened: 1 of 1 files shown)


hiroyuki-komatsu:
We have merged your PR.

PR #1461, opened 2026-04-07
Remove singleton dependency from environ

yukawa:
## Description
As part of our on-going efforts to remove the dependency on `Singleton`, this commit rewrites `base::Environ` without the `Singleton` class dependency.

## Issue IDs
N/A

 - OS: All
 - Steps:


hiroyuki-komatsu:
We have merged your PR.

PR #1463, opened 2026-04-07
Remove `singleton` dependency from `base::Clock`

yukawa:
## Description
As part of our on-going efforts to remove the dependency on `Singleton`, this commit rewrites `base::Clock` without the `Singleton` class dependency.

## Issue IDs

N/A

 - OS: All
 - Steps:

### DIFF
--- a/src/base/clock.cc
+++ b/src/base/clock.cc
@@ -29,11 +29,11 @@
 
 #include "base/clock.h"
 
+#include <atomic>
 #include <ctime>
 
 #include "absl/time/clock.h"
 #include "absl/time/time.h"
-#include "base/singleton.h"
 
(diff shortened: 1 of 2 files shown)


src/base/clock.cc
```diff
@@ -33,7 +33,8 @@
 
 #include "absl/time/clock.h"
 #include "absl/time/time.h"
-#include "base/singleton.h"
+
+#include <atomic>
```
hiroyuki-komatsu: Please add it before `<ctime>`.

```
#include <atomic>
#include <ctime>
```

yukawa: Oops, fixed.

PR #1464, opened 2026-04-07
Remove `singleton` dependency from `base::FileUtil`

yukawa:
## Description
As part of our on-going efforts to remove the dependency on `Singleton`, this commit rewrites `base::FileUtil` without the `Singleton` class dependency.

## Issue IDs
N/A

 - OS: All
 - Steps:

### DIFF
--- a/src/base/file_util.cc
+++ b/src/base/file_util.cc
@@ -29,6 +29,7 @@
 
 #include "base/file_util.h"
 
+#include <atomic>
 #include <cerrno>
 #include <cstddef>
 #include <cstdio>
@@ -53,7 +54,6 @@
 #include "base/file_stream.h"
 #include "base/mmap.h"
 #include "base/port.h"
-#include "base/singleton.h"
 #include "base/strings/zstring_view.h"
 
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


src/base/file_util.cc
```diff
@@ -193,10 +172,11 @@ absl::Status StripWritePreventingAttributesIfExists(zstring_view filename) {
 #endif  // _WIN32
 
 absl::Status FileUtil::CreateDirectory(zstring_view path) {
-  return FileUtilSingleton::Get()->CreateDirectory(path);
-}
+  if (FileUtilInterface* mock = g_mock.load(std::memory_order_acquire);
```
hiroyuki-komatsu: Here's a comment from a reviewer.

The current implementation has a lot of boilerplate, making it hard to update the logic. How about we use a macro like this instead?

```
#define MAYBE_INVOKE_MOCK(method, ...) \
    if (auto* mock = g_mock.load(std::memory_order_acquire)) { \
        return mock->method(__VA_ARGS__); \
    }

absl::Status FileUtil::CreateDirectory(zstring_view path) {
    MAYBE_INVOKE_MOCK(CreateDirectory, path);
```

yukawa: Done in 4f77763dc8382340d6b0541f341992929da6c8b5.

 - https://github.com/google/mozc/compare/e497cf73d22b6307538a30231bcb0d09274a5146..4f77763dc8382340d6b0541f341992929da6c8b5

PR #1465, opened 2026-04-07
Remove unused dependency on `singleton` from `mozcjni`

yukawa:
## Description
`mozcjni.cc` no longer uses `singleton`. Let's remove the unnecessary include and the dependency from `BUILD.bazel`.

## Issue IDs

N/A

 - OS: All
 - Steps:
   1. Confirm all the GitHub Actions still pass


hiroyuki-komatsu:
We have merged your PR.

Issue #1459, opened 2026-04-06
Fully support Per-Monitor DPI v2 in candidate window for Windows


yukawa:
## Description
Candidate windows implemented in `mozc_renderer` currently does not fully support Per-Monitor (V2) DPI. Let's fix remaining issues.

## Steps to reproduce
1. Install Mozc for Windows
2. Open Notepad then enable Mozc.
3. Type something to show the candidate window.
4. Change the display DPI in the system settings.

## Expected behavior
The candidate window is scaled up/down accordingly.

## Actual behavior
The candidate window remains to use the initial DPI.

## Version or commit-id
395f9d7fa9c90e265c953a6deaea2922a368a12a

## Environment
 - OS: Windows 11 25H2

## Additional context
This is one of the major remaining tasks of the following meta bug.
 - https://github.com/google/mozc/issues/831

PR #1467, opened 2026-04-12
Simplify `GetDPIScalingFactor` for Windows

yukawa:
## Description
This is a preparation for fully supporting Per-Monitor DPI v2 in the candidate window for Windows (#1459) without changing the observable behavior.

Unlike the legacy GDI APIs supported anisotropic scaling, which meant that the horizontal and vertical DPI could be different, the latest Per-Monitor DPI v2 only supports isotropic scaling, which means that the horizontal and vertical DPI are always the same. This change simplifies the implementation of `GetDPIScalingFactor` for Windows, as we can now return a single scaling factor instead of separate horizontal and vertical factors.

## Issue IDs

 - OS: Windows 11 25H2
 - Steps:
   1. Change the DPI scaling factor to 2.0
   2. Confirm that Mozc' candidate window is still shown in the same size and position.

### DIFF
--- a/src/renderer/BUILD.bazel
+++ b/src/renderer/BUILD.bazel
@@ -40,7 +40,6 @@ load(
     "BRANDING",
     "MACOS_BUNDLE_ID_PREFIX",
 )
-load("//bazel/win32:build_defs.bzl", "features_gdi")
 
 package(default_visibility = ["//visibility:private"])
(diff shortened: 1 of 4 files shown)


hiroyuki-komatsu:
We have merged your PRs.

PR #1473, opened 2026-04-21
Always evaluate `ServerLauncher::server_program()`

yukawa:
## Description
This is the final preparation step towards switching the installation directory of Mozc for Windows from `"Program Files (x86)"` to `"Program Files"`.

With this commit, `ServerLauncher::server_program()` always return the correct path to `mozc_server.exe` even when its installation directory is moved from `"Program Files (x86)"` to `"Program Files"` while the client app is running. To summarize, remaining steps are:

 1. Release a new version of Mozc for Windows with this change.
 2. Wait for a while until most of the users receive the update.
 3. Release another new version of Mozc for Windows with the installation directory switched to `"Program Files"`.
 4. At this point, we can treat `ServerLauncher::server_program()` as a constant function again.

## Issue IDs

 * https://github.com/google/mozc/issues/1086

 - OS: Windows 11 25H2
 - Steps:
   1. Confirm no observable behavior change at this moment.

### DIFF
--- a/src/client/client_interface.h
+++ b/src/client/client_interface.h
@@ -38,7 +38,6 @@
 
 #include "absl/strings/string_view.h"
 #include "absl/time/time.h"
-#include "base/strings/zstring_view.h"
 #include "ipc/ipc.h"
 #include "protocol/commands.pb.h"
 #include "protocol/config.pb.h"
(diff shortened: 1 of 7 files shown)


src/client/client_interface.h
hiroyuki-komatsu: Please exclude this include.

Note, we are performing refactoring around `zstring_view`. So I probably ask you to rebase this PR when that refactoring is committed before this PR.

hiroyuki-komatsu: Please also update BUILD.bazel too.

yukawa: > Please exclude this include.

Done.

> Please also update BUILD.bazel too.

I think `BUILD.bazel` was already updated to not rely on `zstring_view`. Was I missing something?

hiroyuki-komatsu: That's my misunderstanding. The current `BUILD.bazel` looks good.

PR #1474, opened 2026-04-21
Make `RendererStyleHandler` platform-neutral

yukawa:
## Description
This is a mechanical refactoring commit as a preparation to fully support Per-Monitor DPI v2 in the candidate window for Windows.

Unlike other platforms such as Linux desktop and macOS, RendererStyleHandler has taken care of DPI scaling on Windows in two ways:

 A. `RendererStyleHandler::GetDPIScalingFactor()` is used only on Windows.
 B. `RendererStyleHandler::GetDefaultRendererStyle()` internally takes care of DPI scaling factor for certain style parameters.

Such discrepancy makes it harder to fully support Per-Monitor DPI v2 on Windows because:

 * `RendererStyleHandler::GetDPIScalingFactor()` assumes there is a single DPI scaling factor, which is no longer valid when each monitor may have different DPI scaling factors.
 * We do not want further Windows-specific requirements to be added to `RendererStyleHandler::GetDefaultRendererStyle()`.

With this commit, `RendererStyleHandler` becomes fully platform-neutral, which protects it from upcoming Windows-specific code changes.

There must be no observable behavior change.

## Issue IDs

 - OS: Windows 11 25H2
 - Steps:
   1. Confirm there is no behavior change.

### DIFF
--- a/src/renderer/BUILD.bazel
+++ b/src/renderer/BUILD.bazel
@@ -239,11 +239,7 @@ mozc_cc_library(
     deps = [
         "//base:singleton",
         "//protocol:renderer_cc_proto",
-    ] + mozc_select(
-        windows = [
-            "//bazel/win32:user32",
-        ],
-    ),
+    ],
 )
 
 mozc_cc_test(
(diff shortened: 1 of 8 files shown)


hiroyuki-komatsu:
We have merged your PRs.

PR #1480, opened 2026-04-22
Remove `Singleton` from `Transliterators`

yukawa:
## Description
As part of our on-going efforts to remove the dependency on `Singleton`, this commit rewrites `Transliterators` without the `Singleton`.

This is a purely mechanical change, and the behavior of the code should be identical before and after this change.

## Issue IDs
N/A

 - OS: All
 - Steps:
   1. `bazelisk test //composer:transliterators_test`


hiroyuki-komatsu:
We have merged your PRs.

Issue #831, opened 2023-10-21
Conform Per-Monitor DPI v2 protocol in Windows


yukawa:
## Description
Currently Mozc binaries fall into [System DPI Awareness](https://learn.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows#system-dpi-awareness) category in Windows.

We'd like to make them run under [Per-Monitor v2 Awareness mode](https://learn.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows#system-dpi-awareness).

## Version or commit-id
af209061632ba7dcc64b63bd06eb988974375508

## Environment
 - OS: Windows 11 22H2

### `src/gui/tool/mozc_tool.exe.manifest`
https://github.com/google/mozc/blob/af209061632ba7dcc64b63bd06eb988974375508/src/gui/tool/mozc_tool.exe.manifest#L25-L27
Needs `PerMonitorV2` declaration.
```
      <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2</dpiAwareness>
```

Also, it'd be safer to switch to Qt 6.6 or later version, which contains https://github.com/qt/qtbase/commit/5e0d9a077d28802988182319ae257e9102f0344e.

-----

### `src/renderer/win32/mozc_renderer.exe.manifest`
https://github.com/google/mozc/blob/af209061632ba7dcc64b63bd06eb988974375508/src/renderer/win32/mozc_renderer.exe.manifest#L25-L27
Needs `PerMonitorV2` declaration.
```
      <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2</dpiAwareness>
```

-----

### `src/renderer/win32/indicator_window.cc`

https://github.com/google/mozc/blob/af209061632ba7dcc64b63bd06eb988974375508/src/renderer/win32/indicator_window.cc#L96-L100
We can no longer assume this is a runtime constant.

https://github.com/google/mozc/blob/af209061632ba7dcc64b63bd06eb988974375508/src/renderer/win32/indicator_window.cc#L115-L119
We also need to handle [WM_DPICHANGED](https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-dpichanged).

-----

### `src/renderer/win32/candidate_window.h`

[...]

PR #1481, opened 2026-04-24
Support DPI change in candidate/infolist window

### LINKED ISSUE #1459: Fully support Per-Monitor DPI v2 in candidate window for Windows

yukawa:
## Description
This is part of our ongoing effort to fully conform Per-Monitor DPI v2 protocol in Windows (#831).

With this commit, both the candidate window and the infolist window become fully aware of dynamic DPI changes. The indicator window and icons managed by text input processor (TIP) DLLs remain unchanged.

The key implementation idea is to not rely on `WM_DPICHANGED` message. It is really tempting to implement `WM_DPICHANGED` handler, but it turns out to be redundant for our use cases. Consider the following two cases explained in the` WM_DPICHANGED` message [document](https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-dpichanged).

 1. When the candidate/infolist window is moving into the monitor with a different DPI. In this case, `win32::WindowManager` should be aware of which monitor the candidate/infolist window will be displayed on, and can let the candidate/infolist window know the new DPI without relying on the `WM_DPICHANGED` message.

 2. When the DPI of the current monitor changes while the candidate/infolist window is visible. In this case, the target application window typically re-renders to adapt to the new DPI, which then triggers a fresh layout pass on our candidate/infolist window. As explained above, `win32::WindowManager` always checks the DPI of the target monitor without relying on the `WM_DPICHANGED` message.

Both cases arise because the candidate and infolist windows behave like popups anchored to the cursor location in the target application window; `WM_DPICHANGED` is primarily meaningful for top-level windows that users can drag and move across the monitors.

Other than that, the implementation is rather straightforward: we just need to make sure to keep propagating the DPI information to the text renderer and to update the DPI-dependent resources when the DPI changes.

Closes #1459.

## Issue IDs

 * https://github.com/google/mozc/issues/831

 - OS: Windows 11 25H2
 - Steps:
   1. Install Mozc
   2. Change the text scaling on the monitor to 200%
   3. Open Nodepad
   4. Type something to confirm that Mozc's candidate window uses the scaled font size
   5. Change the text scaling on the monitor to 300%
   6. Type something to confirm that Mozc's candidate window uses the scaled font size

### DIFF
--- a/src/renderer/win32/candidate_window.cc
+++ b/src/renderer/win32/candidate_window.cc
@@ -35,6 +35,7 @@
 #include <wil/resource.h>
 #include <windows.h>
 
+#include <cstdint>
 #include <memory>
 #include <sstream>
 #include <string>
@@ -56,9 +57,6 @@ namespace mozc {
 namespace renderer {
 namespace win32 {
 namespace {
-
-// 96 DPI is the default DPI in Windows.
-constexpr int kDefaultDPI = 96;
 
 // layout size constants in pixel unit in the default DPI.
 constexpr int kIndicatorWidthInDefaultDPI = 4;
@@ -239,12 +237,19 @@ CandidateWindow::CandidateWindow()
       footer_logo_display_size_(0, 0),
       send_command_interface_(nullptr),
       table_layout_(std::make_unique<TableLayout>()),
-      text_renderer_(TextRenderer::Create()),
+      dpi_(::GetDpiForSystem()),
+      text_renderer_(TextRenderer::Create(dpi_)),
       indicator_width_(0),
       metrics_changed_(false),
       mouse_moving_(true) {
+  UpdateDpiDependentResources();
+}
+
+CandidateWindow::~CandidateWindow() = default;
+
+void CandidateWindow::UpdateDpiDependentResources() {
+  const double scale_factor = GetDPIScalingFactor(dpi_);
   double image_scale_factor = 1.0;
-  const double scale_factor = GetDPIScalingFactor();
   if (scale_factor < 1.125) {
     footer_logo_.reset(LoadBitmapFromResource(::GetModuleHandle(nullptr),
                                               IDB_FOOTER_LOGO_COLOR_100));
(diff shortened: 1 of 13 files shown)


src/renderer/win32/candidate_window.cc
```diff
@@ -239,12 +236,19 @@ CandidateWindow::CandidateWindow()
...
-      text_renderer_(TextRenderer::Create()),
+      dpi_(::GetDpiForSystem()),
+      text_renderer_(TextRenderer::Create(dpi_)),
       indicator_width_(0),
       metrics_changed_(false),
       mouse_moving_(true) {
+  UpdateDpiDependentResources();
+}
+
+CandidateWindow::~CandidateWindow() {}
```
hiroyuki-komatsu: Please use '= default' to define a trivial destructor.

https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-equals-default.html

src/renderer/win32/candidate_window.cc
```diff
@@ -276,13 +280,20 @@ CandidateWindow::CandidateWindow()
...
 }
 
-CandidateWindow::~CandidateWindow() {}
-
 LRESULT CandidateWindow::OnCreate(LPCREATESTRUCT create_struct) {
   EnableOrDisableWindowForWorkaround();
   return 0;
 }
 
+void CandidateWindow::UpdateDpi(UINT dpi) {
```
hiroyuki-komatsu: How about using uint32_t?

We can confirm it with static_assert if necessary.

```
static_assert(std::is_same<UINT, uint32_t>::value, "UINT must be uint32_t.");
```

src/renderer/win32/candidate_window.cc
yukawa: Done.

src/renderer/win32/candidate_window.cc
yukawa: Done.

Issue #1484, opened 2026-04-25
Consider using IBus semantic preedit hints


ftake:
## Description
Under Wayland session, the current converting text segment is not highlighted. So users cannot understand which text segment they are converting.

## Steps to reproduce
1. Type 'watashinonamaeha'
2. Press Space

## Expected behavior

A converting text segment is highlighted.

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/0e758034-5ae6-4b0f-a1f4-bb3a36a72bf7" />

## Actual behavior

No text highlight.

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/bb004cc6-d145-4db6-8a65-0833e8a13def" />

## Screenshots
See above

## Version or commit-id
Mozc-2.32.5981.102+24.11.oss

## Environment
 - OS: openSUSE Leap 16.0
 - IMF (for Linux): IBus
 - Related Applications: Kate

## Additional context
A patch to fix this problem is ready for PR.

yukawa:
Thank you for the report. I changed the issue title as what your #1485 is proposing is no more or less than just starting using IBus semantic preedit hints that are available since IBus 1.5.33. I think "Consider using IBus semantic preedit hints" would be more accurate if your goal is to submit #1485.

My point is that it's still possible that the text segment is not highlighted even after submitting your #1485 depending on the target application and other libraries such as IBus / GTK / Qt / Chromium, and so on. Hopefully the new issue title would make it clear that using the new APIs is what Mozc can do at best, and what the user would see is still not fully guaranteed.

Issue #924, opened 2024-05-04
Upgrade reference build environment for Linux/Android to Ubuntu 24.04


yukawa:
## Description
As Ubuntu 24.04 has been released, it's time to start the migration process from Ubuntu 22.04.

Primary reasons are:

 * To switch to newer C++ compiler for better C++20 conformance.
   * https://github.com/google/mozc/issues/783
 * To update Qt6 version
   * In Windows and mac builds we intend to use the latest Qt6 release as expected by upstream, whereas in Linux build we tend to end up having to use the version packaged by the Linux distribution. Although Ubuntu 24.04 gives us [Qt 6.4.2](https://packages.ubuntu.com/noble/qt6-base-dev), which is still quite behind our Window and mac builds that now use Qt 6.7, it'd be way better than Ubuntu 22.04 where we ended up having to work around [QTBUG-86080](https://bugreports.qt.io/browse/QTBUG-86080) as seen in 7b0fc713f090b9dfbe3623f1c16a19e2f7608428.

## Version or commit-id
83641d9adf579e41ab66f625f6699a725b57d671

## Prior work
 * https://github.com/google/mozc/issues/763

yukawa:
Ubuntu 24 for GitHub-hosted runners is now [generally available](https://github.blog/changelog/2024-09-25-actions-new-images-and-ubuntu-latest-changes/).

It should be ready for us to migrate from Ubuntu 22.04 runner to Ubuntu 24.04 runner.

Issue #763, opened 2023-06-17
Upgrade the linux GitHub action to Ubuntu 22.04


yuryu:
The current version is 20.04, which is more than three years old. Upgrade it to 22.04 to keep up with the support lifecycle and use the latest compiler.

https://github.com/google/mozc/actions/workflows/linux.yaml

PR #1485, opened 2026-04-25
WIP: Fix preedit decoration in ibus-mozc under Wayland

ftake:
## Description
Wayland protocols do not support specifying arbitrary RGB colors for preedit styling. This commit switchs preedidt decoration to IBus semantic preedit hints (available since 1.5.33) so that decoration works correctly under Wayland.

## Issue IDs
#1484

 - OS: Linux, Plasma Wayland
 - Steps:
   1. Open Kate
   2. Type Watashinonamaeha
   3. Type Space
   4. Check the first segment is highlighted

## Additional context

* This fix requires newer IBus (1.5.33). If necessary, we can keep some codes for backward compatibility.
* This change affects X11 session. The highlight color will be changed from `#D1EAFF` to IBus's default.

### DIFF
--- a/src/unix/ibus/preedit_handler.cc
+++ b/src/unix/ibus/preedit_handler.cc
@@ -52,8 +52,23 @@ IbusTextWrapper ComposePreeditText(const commands::Preedit& preedit) {
   int end = 0;
   for (int i = 0; i < preedit.segment_size(); ++i) {
     const commands::Preedit::Segment& segment = preedit.segment(i);
+#if IBUS_CHECK_VERSION(1, 5, 33)
+    // IBus's semantic pre-edit style name is since 1.5.33, which is necessary
+    // to use pre-edit style under Wayland session.
+    IBusAttrPreedit hint = IBUS_ATTR_PREEDIT_WHOLE;
+#else  // libibus (>=1.5.33)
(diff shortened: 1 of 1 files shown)


yukawa:
> This fix requires newer IBus (1.5.33). If necessary, we can keep some codes for backward compatibility.

Our CI relies on Ubuntu 24.04, where IBus 1.5.29 seems to be used. Does your change break our CI? If yes, I think the answer is clear and we need to keep the backward compatibility.

https://github.com/google/mozc/blob/a2303d2712bcbdbd3368c900e5e5076e5e95c35c/.github/workflows/linux.yaml#L16-L19

ftake:
> Does your change break our CI?

Yes...
Is there any plan to update the CI environment to Ubuntu 26.04?

yukawa:
> Is there any plan to update the CI environment to Ubuntu 26.04?

Yes, we are going to update to Ubuntu 26.04 when it becomes available. See the following tasks when we updated our CI Ubuntu version in the past.
 - https://github.com/google/mozc/issues/924
 - https://github.com/google/mozc/issues/763

That said chances are high that GitHub will not release new Ubuntu version soon. As for Ubuntu 24.04, it was September 25 when it became GA.
 - https://github.blog/changelog/2024-09-25-actions-new-images-and-ubuntu-latest-changes/

I think keeping backward compatibility is a realistic path forward.

PR #1489, opened 2026-04-26
Remove redundant build options from instructions

yukawa:
## Description
With recent commits (51ee0a77d00709ebbb977ac723ce131ed4d1fbe2)(5d460978ef0ba5509651016168930de02a999386), we can remove redundant build options from our build instructions and GitHub Actions workflows.

Key changes are:
 - Removed `--config oss_{linux,macos,windows}` from build instructions and workflows as `--enable_platform_specific_config` option automatically now sets them.
 - Removed `MOZC_QT_PATH` environment variable from build instructions and workflows from macOS build instructions and workflows as its default value is already set to `third_party/qt`, where `build_qt.py` deploys Qt by default.

There must be no functional change in the final artifacts, but the instructions and workflows are now simplified.

## Issue IDs
N/A

 - OS: All
 - Steps:
   1. GitHub Actions still pass

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -36,7 +36,7 @@ jobs:
       - name: bazel build
         working-directory: ./src
         run: |
-          bazelisk build --config oss_linux package --config release_build
+          bazelisk build package --config release_build
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v6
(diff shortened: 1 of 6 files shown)


hiroyuki-komatsu:
Thank you for the PR.
Since the latest tagged version still requires those options, we will commit this change after or just before the next tagging.

PR #1492, opened 2026-04-27
Fix MSVC extension warnings in `WinAPITestHelper`

yukawa:
## Description
The implicit function-to-object pointer conversions triggered Clang's `-Wmicrosoft-cast` warning. Let's use the alias to a generic function-pointer type `void (*)()` and `reinterpret_cast` at the two boundaries that produce specific function-pointer types.

`WinAPITestHelper` is used only for unit tests, so there must be no risk in the production code.

## Issue IDs
N/A

 - OS: Windows
 - Steps:
   1. Confirm all tests still pass

### DIFF
--- a/src/base/win32/win_api_test_helper.cc
+++ b/src/base/win32/win_api_test_helper.cc
@@ -114,7 +114,8 @@ class HookTargetInfo {
         continue;
       }
       const FunctionPointer original_proc_address =
-          ::GetProcAddress(module_handle, request.proc_name.c_str());
+          reinterpret_cast<FunctionPointer>(
+              ::GetProcAddress(module_handle, request.proc_name.c_str()));
       if (original_proc_address == nullptr) {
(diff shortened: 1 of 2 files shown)


src/base/win32/win_api_test_helper.cc
```diff
@@ -113,8 +113,8 @@ class HookTargetInfo {
         LOG(FATAL) << "GetModuleHandleExA failed. error = " << error;
         continue;
       }
-      const FunctionPointer original_proc_address =
-          ::GetProcAddress(module_handle, request.proc_name.c_str());
+      const FunctionPointer original_proc_address = reinterpret_cast<FunctionPointer>(
```
hiroyuki-komatsu: Please wrap the line in 80 chars.

yukawa: Done.

hiroyuki-komatsu:
We have merged your PRs.

PR #1493, opened 2026-04-28
Introduce `ConstInitImmutableString` and `ConstInitMutableString`

yukawa:
## Description

These are utility classes for global strings whose value is known only at runtime. They are motivated by the on-going effort to remove `Singleton<T>`, which has surfaced several places in `system_util.cc` (directory caches, user-SID lookup, program-invocation-name holder) where the singleton wraps a string lookup that must run **after** the surrounding modules are ready.

A function-local `static std::string` would be the obvious replacement, but reintroduces two issues that `Singleton<T>` callers have learnt to avoid:

  * On Windows, function-local static initialization is not safe to invoke from `DllMain` because the compiler-generated guard can take locks that interact badly with the Loader Lock.
  * The resulting destructor runs at process exit in unspecified order relative to other globals.

Both classes address these by being trivially destructible (values that don't fit inline live in an intentionally-leaked heap buffer) and by keeping all foreign code outside the publish lock.

`ConstInitImmutableString` is the publish-once, read-many variant. It takes an `IdempotentInitializer` function pointer that is invoked lock-free on the first `GetOrInit()`, so the class is safe to use from `DllMain` even when the initializer itself acquires the Loader Lock. Subsequent reads return a `string_view` over stable, NUL-terminated storage on a lock-free fast path:

```cpp
#include "base/strings/const_init_immutable_string.h"

constinit mozc::ConstInitImmutableString<256> kProgramFilesX86(
    []() -> std::string { return ComputeProgramFilesX86Path(); });

absl::string_view path = kProgramFilesX86.GetOrInit();
```

`ConstInitMutableString` is the set-many, snapshot-read variant. Set atomically replaces the stored value; Get returns a `std::string` snapshot under a spinlock-protected copy. Empty doubles as the "never set" sentinel; callers that want a lazy default observe empty and Set it themselves:

```cpp
#include "base/strings/const_init_mutable_string.h"

constinit mozc::ConstInitMutableString<256> g_invocation_name;

void Init(absl::string_view name) { g_invocation_name.Set(name); }
std::string GetName()             { return g_invocation_name.Get(); }
```

Implementation notes:

  * Only the publish step is serialized, by a small `atomic_flag` spinlock with PAUSE / YIELD hints. Heap allocation for the fallback buffer is staged before the spinlock is taken, so the critical section is bounded by `fixed_array_size` and a throwing `bad_alloc` cannot leave the spinlock latched.
  * Constructors are `consteval`, so instances must be constant-initialized.
  * Common helpers (spin hint, RAII spinlock guard, heap-fallback staging and commit) live in `base/strings/internal/const_init_string_helpers.{h,cc}`. Templates that operate on `CharT` data are explicitly instantiated for `char` and `wchar_t` so they live in one TU.

## Issue IDs
N/A

 - OS: All
 - Steps:
   1. `bazelisk test //base/strings/...  --build_tests_only`
 
## Additional context

### DIFF
--- a/src/base/strings/const_init_immutable_string.h
+++ b/src/base/strings/const_init_immutable_string.h
@@ -1,0 +1,142 @@
+// Copyright 2010-2021, Google Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef MOZC_BASE_STRINGS_CONST_INIT_IMMUTABLE_STRING_H_
+#define MOZC_BASE_STRINGS_CONST_INIT_IMMUTABLE_STRING_H_
+
+#include <array>
+#include <atomic>
+#include <cstddef>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <string_view>
+#include <type_traits>
+
+#include "base/strings/internal/const_init_string_helpers.h"
+
+namespace mozc {
+
+// A utility class to deal with constant global strings whose value is known
+// only at runtime. It has the following capabilities:
+// 1. It allows the library users to lazily initialize the string by calling
+//    `GetOrInit()` only after it becomes ready, e.g. only after dependent
+//    modules are fully loaded.
+// 2. It is thread-safe, meaning that multiple threads can call `GetOrInit()`
+//    concurrently without causing data race, with an assumption that the
+//    `idempotent_initializer` is idempotent (i.e., it always returns the same
+//    value when called multiple times) and thread-safe.
+// 3. It guarantees that the string is null-terminated.
+// 4. It is trivially destructible, which means it can be used in static
+//    storage duration objects without causing destructor order issues, with a
+//    caveat that it may leak memory if the string is larger than the fixed
+//    array size provided by the template parameter.
+//
+// Once the string is published it cannot be replaced. See
+// `ConstInitMutableString` for a variant that additionally supports
+// thread-safe `Set()`.
+template <size_t fixed_array_size,
+          const_init_string_internal::SupportedChar CharT = char>
+class ConstInitImmutableString {
+ public:
+  using StringT = std::basic_string<CharT>;
+  using StringViewT = std::basic_string_view<CharT>;
+
+  // A data initializer that is guaranteed to return the same value no matter
+  // how many times it is called. It must also be reentrant and safe to call
+  // from multiple threads concurrently.
+  using IdempotentInitializer = std::add_pointer_t<StringT()>;
+
+  ConstInitImmutableString() = delete;
+  ~ConstInitImmutableString() = default;
+
+  consteval explicit ConstInitImmutableString(
+      IdempotentInitializer idempotent_initializer)
+      : idempotent_initializer_(idempotent_initializer) {}
+
+  [[nodiscard]] StringViewT GetOrInit() {
+    // Fast path: result already published.
+    if (const CharT* ptr = result_ptr_.load(std::memory_order::acquire))
+        [[likely]] {
+      return StringViewT(ptr, result_size_);
+    }
+
+    // The initializer is called outside any lock so that no foreign code
+    // runs under a lock held by this class. This avoids classic deadlock
+    // scenarios such as Win32 Loader Lock recursion when `GetOrInit()` is
+    // called from `DllMain` and the initializer internally invokes
+    // `LoadLibrary`. Racing threads may each call it; per the
+    // `IdempotentInitializer` contract that is acceptable.
+    // https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-best-practices#deadlocks-caused-by-lock-order-inversion
+    const StringT value = idempotent_initializer_();
+
+    // Stage any heap fallback outside the spinlock so that the critical
+    // section is bounded by `fixed_array_size` and a throwing
+    // allocation cannot leave the spinlock held.
+    std::unique_ptr<CharT[]> heap_fallback =
+        const_init_string_internal::StageHeapFallback<CharT>(value,
+                                                             std::size(value_));
+
+    std::lock_guard<const_init_string_internal::TrivialMicroSpinLock> l(
+        committing_);
+    if (const CharT* ptr = result_ptr_.load(std::memory_order::relaxed)) {
+      // Another thread already published; staged `heap_fallback` is
+      // freed by its destructor when this scope unwinds.
+      return StringViewT(ptr, result_size_);
+    }
+
+    // Winner: any staged heap buffer is intentionally leaked for the
+    // remainder of the process so that `ConstInitImmutableString` itself
+    // remains trivially destructible.
+    CharT* dest = const_init_string_internal::CommitStagedValue<CharT>(
+        value, heap_fallback, value_.data());
+    result_size_ = value.size();
+    result_ptr_.store(dest, std::memory_order::release);
+    return StringViewT(dest, value.size());
+  }
+
+ private:
+  std::atomic<const CharT*> result_ptr_ = nullptr;
+  size_t result_size_ = 0;
+  const_init_string_internal::TrivialMicroSpinLock committing_;
+  std::array<CharT, fixed_array_size> value_ = {};
+  const IdempotentInitializer idempotent_initializer_;
+};
+
+// Verify the trivial destructibility contract for both supported `CharT`
+// instantiations so misuse (e.g. accidentally adding a non-trivial member)
+// is caught at compile time even if no instance is constructed.
+static_assert(
+    std::is_trivially_destructible_v<ConstInitImmutableString<1, char>>);
+static_assert(
+    std::is_trivially_destructible_v<ConstInitImmutableString<1, wchar_t>>);
+
+}  // namespace mozc
+
+#endif  // MOZC_BASE_STRINGS_CONST_INIT_IMMUTABLE_STRING_H_
--- a/src/base/strings/const_init_immutable_string_test.cc
+++ b/src/base/strings/const_init_immutable_string_test.cc
@@ -1,0 +1,78 @@
+// Copyright 2010-2021, Google Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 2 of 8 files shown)


src/base/strings/internal/const_init_string_helpers.cc
```diff
@@ -0,0 +1,80 @@
...
+#include <concepts>
+#include <cstddef>
+#include <cstring>
+#include <memory>
+
+namespace mozc::const_init_string_internal {
+
+template <typename CharT>
+requires std::same_as<CharT, char> || std::same_as<CharT, wchar_t>
+std::unique_ptr<CharT[]> StageHeapFallback(const CharT* data, size_t size,
```
hiroyuki-komatsu: How about using `concept` to reuse `requires`.

```
template <typename T>
concept SupportedChar = std::same_as<T, char> || std::same_as<T, wchar_t>;

template <size_t fixed_array_size, SupportedChar CharT = char>
class ConstInitImmutableString { ... };
```

src/base/strings/internal/const_init_string_helpers.cc
```diff
@@ -0,0 +1,80 @@
...
+                                           size_t inline_capacity) {
+  const size_t size_with_null = size + 1;
+  if (size_with_null <= inline_capacity) {
+    return nullptr;
+  }
+  // `new` (default-init) avoids the zero-fill that `make_unique` would
+  // perform before `memcpy` overwrites the buffer.
+  std::unique_ptr<CharT[]> heap(new CharT[size_with_null]);
```
hiroyuki-komatsu: If we can use `std::make_unique` here, I'd prefer using it although the performance is slightly negative.

```
auto heap = std::make_unique<CharT[]>(size_with_null);
```

src/base/strings/internal/const_init_string_helpers.cc
```diff
@@ -0,0 +1,80 @@
...
+  std::memcpy(heap.get(), data, size * sizeof(CharT));
```
hiroyuki-komatsu: How about `std::copy_n` ?

```
std::copy_n(data, size, heap.get());
```

src/base/strings/const_init_mutable_string.h
```diff
@@ -0,0 +1,139 @@
...
+#include <concepts>
+#include <cstddef>
+#include <cstring>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <string_view>
+#include <type_traits>
+
+#include "base/strings/const_init_immutable_string.h"
```
hiroyuki-komatsu: "const_init_string_helpers.h" ?

src/base/strings/const_init_mutable_string.h
```diff
@@ -0,0 +1,139 @@
...
+requires std::same_as<CharT, char> || std::same_as<CharT, wchar_t>
+void ConstInitMutableString<fixed_array_size, CharT>::Set(StringViewT value) {
+  // Stage any heap fallback outside the spinlock so the critical section
+  // is bounded by `fixed_array_size` and a throwing `bad_alloc` cannot
+  // leave the spinlock held.
+  std::unique_ptr<CharT[]> heap_fallback =
+      const_init_string_internal::StageHeapFallback<CharT>(
+          value.data(), value.size(), std::size(value_));
+
+  CharT* old_heap = nullptr;
```
hiroyuki-komatsu: How about using `std::unique_ptr` here?

```
std::unique_ptr<CharT[]> old_heap;
{
  std::lock_guard<const_init_string_internal::SpinLock> l(committing_);
  if (current_ptr_ != nullptr && current_ptr_ != value_.data()) {
    old_heap.reset(current_ptr_);
  }
  // ...
}
```

src/base/strings/internal/const_init_string_helpers.h
```diff
@@ -0,0 +1,124 @@
...
+// character type is a link error by design.
+
+// Stages a copy of `[data, data + size)` for publication. If `size + 1`
+// characters (including a NUL terminator) fit in `inline_capacity`,
+// returns a null pointer and the caller is expected to commit to its
+// own inline buffer; otherwise returns a fresh, NUL-terminated heap
+// copy.
+template <typename CharT>
+requires std::same_as<CharT, char> || std::same_as<CharT, wchar_t>
+std::unique_ptr<CharT[]> StageHeapFallback(const CharT* data, size_t size,
```
hiroyuki-komatsu: How about using `StringViewT` (or `std::basic_string_view<CharT>`)?

```
std::unique_ptr<CharT[]> StageHeapFallback(const StringViewT value, size_t inline_capacity);
```

src/base/strings/const_init_mutable_string.h
```diff
@@ -0,0 +1,139 @@
...
+  // The `lock_guard` releases the lock if the `StringT` allocation
+  // here throws.
+  std::lock_guard<const_init_string_internal::SpinLock> l(committing_);
+  return StringT(current_ptr_, current_size_);
+}
+
+// Verify the trivial destructibility contract for both supported `CharT`
+// instantiations so misuse (e.g. accidentally adding a non-trivial member)
+// is caught at compile time even if no instance is constructed.
+static_assert(std::is_trivially_destructible_v<ConstInitMutableString<1, char>>);
```
hiroyuki-komatsu: Please wrap in 80 chars.

src/base/strings/const_init_immutable_string_test.cc
```diff
@@ -0,0 +1,82 @@
...
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "testing/gunit.h"
+
+namespace mozc::strings {
+namespace {
+
+constinit ConstInitImmutableString<256> g_value_simple_init([]() -> std::string {
```
hiroyuki-komatsu: Please wrap in 80 chars.

src/base/strings/const_init_immutable_string_test.cc
```diff
@@ -0,0 +1,82 @@
...
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "base/strings/const_init_immutable_string.h"
+
+#include <atomic>
```
hiroyuki-komatsu: `<atomic>` is not used.

src/base/strings/const_init_immutable_string_test.cc
```diff
@@ -0,0 +1,82 @@
...
+#include <string>
+#include <string_view>
+#include <vector>
```
hiroyuki-komatsu: `<vector>` is not used.

src/base/strings/const_init_immutable_string_test.cc
```diff
@@ -0,0 +1,82 @@
...
+
+TEST(ConstInitImmutableStringTest, HeapFallback) {
+  const auto value = g_value_overflow.GetOrInit();
+  EXPECT_EQ(value, kLongValue);
+  EXPECT_EQ(value.data()[value.size()], '\0');
+  // Repeated calls return the same pointer (stable storage).
+  EXPECT_EQ(g_value_overflow.GetOrInit().data(), value.data());
+}
+
+constinit ConstInitImmutableString<256, wchar_t> g_value_wide([]() -> std::wstring {
```
hiroyuki-komatsu: Please wrap in 80 chars.

src/base/strings/const_init_immutable_string.h
```diff
@@ -0,0 +1,140 @@
...
+requires std::same_as<CharT, char> || std::same_as<CharT, wchar_t>
+class ConstInitImmutableString {
+ public:
+  using StringT = std::basic_string<CharT>;
+  using StringViewT = std::basic_string_view<CharT>;
+
+  // A data initializer that is guaranteed to return the same value no matter
+  // how many times it is called. It must also be reentrant and safe to call
+  // from multiple threads concurrently.
+  using IdempotentInitializer = StringT (*)();
```
hiroyuki-komatsu: How about using `std::add_pointer_t`?

```
using IdempotentInitializer = std::add_pointer_t<StringT()>;
```

src/base/strings/const_init_immutable_string.h
```diff
@@ -0,0 +1,140 @@
...
+  size_t result_size_ = 0;
+  const_init_string_internal::SpinLock committing_;
+  std::array<CharT, fixed_array_size> value_ = {};
+  const IdempotentInitializer idempotent_initializer_;
+};
+
+// Verify the trivial destructibility contract for both supported `CharT`
+// instantiations so misuse (e.g. accidentally adding a non-trivial member)
+// is caught at compile time even if no instance is constructed.
+static_assert(std::is_trivially_destructible_v<ConstInitImmutableString<1, char>>);
```
hiroyuki-komatsu: Please wrap in 80 chars.

src/base/strings/const_init_immutable_string.h
```diff
@@ -0,0 +1,140 @@
...
+
+    std::lock_guard<const_init_string_internal::SpinLock> l(committing_);
+    if (const CharT* ptr = result_ptr_.load(std::memory_order::relaxed)) {
+      // Another thread already published; staged `heap_fallback` is
+      // freed by its destructor when this scope unwinds.
+      return StringViewT(ptr, result_size_);
+    }
+
+    // Winner: any staged heap buffer is intentionally leaked for the
+    // remainder of the process so that `ConstInitImmutableString` itself remains
```
hiroyuki-komatsu: Please wrap in 80 chars.

hiroyuki-komatsu (commented):
Thank you for the PR!

src/base/strings/internal/const_init_string_helpers.cc
yukawa: Done.

src/base/strings/const_init_mutable_string.h
yukawa: Done.

src/base/strings/internal/const_init_string_helpers.h
yukawa: Switched to `std::basic_string_view<CharT>`. Done.

src/base/strings/const_init_mutable_string.h
yukawa: Done.

src/base/strings/const_init_mutable_string.h
yukawa: Oops, fixed.

src/base/strings/internal/const_init_string_helpers.cc
yukawa: Done.

src/base/strings/internal/const_init_string_helpers.cc
yukawa: Switched to `std::make_unique_for_overwrite`. Let me know if you prefer `std::make_unique`.

src/base/strings/const_init_immutable_string_test.cc
yukawa: Done.

src/base/strings/const_init_immutable_string_test.cc
yukawa: Done.

src/base/strings/const_init_immutable_string_test.cc
yukawa: Done.

src/base/strings/const_init_immutable_string.h
yukawa: Done.

src/base/strings/const_init_immutable_string.h
yukawa: Done.

src/base/strings/const_init_immutable_string.h
yukawa: Done.

src/base/strings/const_init_immutable_string_test.cc
yukawa: Done.

yukawa:
Done. Let me know if you prefer `std::make_unique` over `std::make_unique_for_overwrite`.

src/base/strings/internal/const_init_string_helpers.h
```diff
@@ -0,0 +1,129 @@
...
+#include <atomic>
+#include <concepts>
+#include <cstddef>
+#include <memory>
+#include <string_view>
+#include <type_traits>
+
+#if defined(_MSC_VER)
+#include <intrin.h>  // _mm_pause, __yield
+#endif
```
hiroyuki-komatsu: Please add a comment to endif. It's required by our internal mechanism.
Here's an example. Other comment is also fine here.

```
#if defined(_MSC_VER)
#include <intrin.h>  // _mm_pause, __yield
#endif  // _MSC_VER
```

src/base/strings/internal/const_init_string_helpers.h
```diff
@@ -0,0 +1,129 @@
...
+#if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))
+  _mm_pause();
+#elif defined(_MSC_VER) && defined(_M_ARM64)
+  __yield();
+#elif (defined(__GNUC__) || defined(__clang__)) && \
+    (defined(__x86_64__) || defined(__i386__))
+  __builtin_ia32_pause();
+#elif (defined(__GNUC__) || defined(__clang__)) && defined(__aarch64__)
+  __asm__ __volatile__("yield" ::: "memory");
+#endif
```
hiroyuki-komatsu: Please add a comment to endif. It's required by our internal mechanism.

Here's an example. Other comment is also fine here.

```
#endif  // (__GNUC__ || __clang__) && __aarch64__
```

src/base/strings/internal/const_init_string_helpers.h
yukawa: Done.

src/base/strings/internal/const_init_string_helpers.h
yukawa: Done.

src/base/strings/internal/const_init_string_helpers.h
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// `atomic_flag`-based spinlock satisfying the `BasicLockable` named
+// requirement, so it composes with `std::lock_guard`, `std::scoped_lock`
+// and `std::unique_lock`.
+//
+// Trivially destructible and `constexpr`-default-constructible (via
+// `std::atomic_flag`'s C++20 default ctor), so instances may live as
+// `mutable` members of `consteval`-constructed `constinit` globals.
+class SpinLock {
```
hiroyuki-komatsu: Is it possible to use absl/base/internal/spinlock.h?

If it's not possible, could you add a comment about it?

yukawa: I looked at `absl/base/internal/spinlock.h` and my conclusion is that we cannot (and should not) use it.

Left comments there. Also renamed it to `const_init_string_internal::TrivialMicroSpinLock` to clarify that we are trying to have a common implementation details specifically for const-init string just to avoid code duplicate, rather than trying to implement a general purpose spin lock class.

Detailed diffs can be found in http://github.com/google/mozc/compare/c62288f632e08d7998d1dee35a1ca9936990938e..1797c00cbf20f302df043d5e34bc4f0ef2c1710a

yukawa: @hiroyuki-komatsu Also started the following discussion thread related to this topic.

 * https://github.com/google/mozc/discussions/1511

yukawa: I've uploaded another variant of the implementation that relies on `std::atomic_flag::wait()` instead of custom implementation of the spinlock with a fallback `absl::Mutex`-based implementation for platforms where `std::atomic_flag::wait()` is disallowed.

 * https://github.com/google/mozc/pull/1512

yukawa:
I've uploaded another variant of the implementation that relies on `std::atomic_flag::wait()` instead of custom implementation of the spinlock with a fallback `absl::Mutex`-based implementation for platforms where `std::atomic_flag::wait()` is disallowed.

 * https://github.com/google/mozc/pull/1512

yukawa:
Thank you for the detailed review. As we talked offline, I decided to abandon this commit for now as we have at least removed `Singleton<T>` from `system_util.cc` as follows:

 * https://github.com/google/mozc/pull/1523

Maybe we can revisit here later (e.g. creating Win32-specifc optimization somewhere in `src/base/win32/win_util.cc`), but for now let's close this for now to minimize the number of open pull requests.

PR #1495, opened 2026-04-29
Allow building Mozc with Visual Studio 2026

### LINKED ISSUE #1487: Support building Mozc with Visual Studio 2026
## Description
As GitHub-hosted runners with Visual Studio 2026 is scheduled to become generally available [on May 4, 2026](https://github.blog/changelog/2026-02-05-github-actions-early-february-2026-updates/#windows-server-2025-with-visual-studio-2026-image-now-available-for-github-hosted-runners), it's time to start supporting building Mozc with Visual Studio 2026.

Before we fully migrate to Visual Studio 2026 and stop supporting Visual Studio 2022, we are going to support the following 3 scenarios:

| Visual Studio 2022 | Visual Studio 2026 | Actually used toolchain |
| ------------------ | ------------------ | ----------------------- |
| ✅ Installed      | Not installed      | Visual Studio 2022      |
| Not installed     | ✅ Installed        | Visual Studio 2026      |
| ✅ Installed      | ✅ Installed          | Visual Studio 2022      |

The current plan is to prioritize Visual Studio 2022 over Visual Studio 2026. We can later introduce some opt-in mechanism to use Visual Studio 2026 over Visual Studio 2022.

## Steps to reproduce
1. Make sure Visual Studio 2026 is installed without Visual Studio 2022.
2. `python build_tools/build_qt.py --release --confirm_license`
3. `bazelisk build --config release_build package`

## Expected behavior
 * You can build Mozc with Visual Studio 2026 at the step 3
 * The installer bundles MSVC CRT for Visual Studio 2026.

## Version or commit-id
 * a2303d2712bcbdbd3368c900e5e5076e5e95c35c

## Environment
 - OS: Windows 11 25H2

## Additional Context
Here are past discussions:
 * https://github.com/google/mozc/discussions/1408
 * https://github.com/google/mozc/pull/1407

yukawa:
## Description
This makes Mozc build on machines where only Visual Studio 2026 is installed, while keeping Visual Studio 2022 as the default toolchain on machines where VS 2022 is available. GitHub Actions runners stay on VS 2022 for the time being; explicit opt-in to VS 2026 when both versions are present can be added in a subsequent commit.

Three coordinated changes make this work:

1. Have `vs_util.py` call `vswhere` twice when the first call returns nothing: `[17,18)` first to find VS 2022, then `[18,19)` to find VS 2026. Combined with `-latest`, the resulting matrix is:
   * Only VS 2022 installed -> picks VS 2022 (no behavior change).
   * Only VS 2026 installed -> picks VS 2026.
   * Both installed         -> picks VS 2022 (no behavior change).

2. Pass `BAZEL_VC` into the installer `genrule`'s command line as `--vs_install_dir`, derive `vcvarsall.bat` from it inside
   `build_installer.py`, and feed it to `vs_util` as the path hint. Add `--action_env=BAZEL_VC` under `config:windows_env` so Bazel propagates the value into action environments. Without this, `build_installer.py` would run its own `vswhere` lookup and could pick a different VS than Bazel did, especially on machines with multiple installs.

3. Pick the CRT redist subfolder dynamically. The subfolder is named after the platform toolset, not the MSVC compiler version: VS 2022 stayed on toolset v143 even after MSVC ticked into the 14.4x range (to avoid breaking existing project files / CI pipelines), so its redist lives under `Microsoft.VC143.CRT`. VS 2026 jumped the toolset to v145 (matching MSVC 14.50), skipping v144 entirely; its redist lives under `Microsoft.VC145.CRT`. Read `VCTOOLSVERSION` (which `vcvarsall.bat` exports next to `VCTOOLSREDISTDIR`) and pick the v145 subfolder when the minor component is >=50, otherwise v143.

Closes #1487.

## Issue IDs

 * https://github.com/google/mozc/issues/1487

 - OS: Windows 11 25H2
 - Steps:
 - Steps:
   1. Install both Visual Studio 2022 and Visual Studio 2026
   2. `python build_tools/build_qt.py --release --confirm_license`
   3. `bazelisk build --config release_build package`
   5. Confirm that Mozc still is built with Visual Studio 2022

### DIFF
--- a/docs/build_mozc_in_windows.md
+++ b/docs/build_mozc_in_windows.md
@@ -45,7 +45,12 @@ Building Mozc on Windows requires the following software.
 *   `.NET 6` or later (for `dotnet` command).
 *   [Bazelisk](https://github.com/bazelbuild/bazelisk)
 
-> [!NOTE] Bazelisk is a wrapper of [Bazel](https://bazel.build) that allows you
+> [!TIP]
+> Visual Studio 2026 Community Edition is also supported to build Mozc. When
+> both VS 2022 and 2026 are installed, VS 2022 will be used.
+
+> [!NOTE]
+> Bazelisk is a wrapper of [Bazel](https://bazel.build) that allows you
 > to use a specific version of Bazel.
 
 ### Download the repository from GitHub
--- a/src/.bazelrc
+++ b/src/.bazelrc
@@ -39,6 +39,7 @@ build:windows_env --extra_toolchains=@local_config_cc//:cc-toolchain-x64_windows
 build:windows_env --extra_toolchains=@local_config_cc//:cc-toolchain-x64_x86_windows-clang-cl
 build:windows_env --host_platform=//:host-windows-clang-cl
 build:windows_env --repo_env=BAZEL_WIN32_WINNT=/D_WIN32_WINNT=0x0A00
+build:windows_env --action_env=BAZEL_VC
 
 ## Compiler options
 common:linux_env   --config=compiler_gcc_like
--- a/src/build_tools/vs_util.py
+++ b/src/build_tools/vs_util.py
@@ -77,7 +77,7 @@ def get_vcvarsall(
         r' --vcvarsall_path=C:\VS\VC\Auxiliary\Build\vcvarsall.bat'
     )
 
-  cmd = [
+  base_cmd = [
       str(vswhere_path),
       '-products',
       'Microsoft.VisualStudio.Product.Enterprise',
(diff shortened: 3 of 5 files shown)


hiroyuki-komatsu:
Could you please keep VS2022 as the default if both versions are available at this moment?

yukawa:
> Could you please keep VS2022 as the default if both versions are available at this moment?

Done. I also dropped the opt-in mechanism to use VS2026 when both VS2022 and VS2026 are installed. It can be implemented in a subsequent commit if necessary (e.g. setting up a continuous build in GitHub Actions). Anyway external contributors can now uninstall VS2022 if Mozc is the last blocker to do so.

PR #1497, opened 2026-05-01
Remove `Singleton` from `PasswordManager`

yukawa:
## Description
As part of our ongoing efforts to remove the dependency on `Singleton`, this commit rewrites `base/password_manager.cc` without the `Singleton` class dependency. There is no observable behavioral change in production.

## Issue IDs
N/A

 - OS: All
 - Steps:
   1. `bazelisk test //base:encryptor_test`


src/base/password_manager.cc
```diff
@@ -39,15 +39,13 @@
 #include <cstdlib>
 #include <string>
 
-#include "absl/log/check.h"
 #include "absl/log/log.h"
```
hiroyuki-komatsu, line 43: Please include absl/base/const_init.h

```
#include "absl/base/const_init.h"
#include "absl/status/status.h"
```

yukawa: Done.

PR #1507, opened 2026-05-09
Emacs: Fix candidate overlay collapse when tab-line is enabled

kkatsuyuki:
## Description
On Emacs, candidate overlay collapses with the tab line shown. This pull-request fixes the problem.

## Issue IDs
None.

 - My environment : Arch Linux 6.18.7, emacs 30.2
 - Steps:
   1. Enable mozc on Emacs. Set `mozc-candidate-style` overlay (default).
   2. Enable tab-line on Emacs, ex. `M-x global-tab-line-mode`.
   3. Input Japanese characters and show completion candidates.
   4. Confirm the candidate overlay display.

## Additional context
Mozc.el seems to support emacs >= 22, but if it drops support of emacs < 27,  the code becomes simpler.

### DIFF
--- a/src/unix/emacs/mozc.el
+++ b/src/unix/emacs/mozc.el
@@ -662,9 +662,11 @@ This hack could be moved to mozc-posn-at-x-y in a future version."
            (+ y (window-tab-line-height) (window-header-line-height)))
           ((or (null header-line-format) (<= emacs-major-version 23))
            y)
-          (t (+ y (or mozc-cached-header-line-height
-                      (setq mozc-cached-header-line-height
-                            (mozc-header-line-height)) 0))))))
+          (t
+           (+ y (or mozc-cached-header-line-height
+                    (setq mozc-cached-header-line-height
+                          (mozc-header-line-height))
+                    0))))))
 
 (defsubst mozc-window-width (&optional window)
   "Return the width of WINDOW in pixel.
--- a/src/unix/emacs/mozc.el
+++ b/src/unix/emacs/mozc.el
@@ -643,7 +643,7 @@ In a word, this function is a fixed version of `posn-at-point'."
   "Return the y coordinate in POSITION.
 
 This function returns y offset from the top of the buffer area including
-the header line.  This definition could be changed in future.
+the header line and tab line.  This definition could be changed in future.
 
 Note: On Emacs 22 and 23, y offset, returned by `posn-at-point' and taken
 by `posn-at-x-y', is relative to the top of the buffer area including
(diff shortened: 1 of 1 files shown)


src/unix/emacs/mozc.el
```diff
@@ -652,15 +652,19 @@ However, on Emacs 24, y offset returned by `posn-at-point' is relative to
...
 This function fixes the asymmetry between them on Emacs 24 and later versions.
 This hack could be moved to mozc-posn-at-x-y in a future version."
   (let ((y (cdr (posn-x-y position))))
-    (if (or (null header-line-format) (<= emacs-major-version 23))
-        y
-      (+ y (or mozc-cached-header-line-height
-               (setq mozc-cached-header-line-height (mozc-header-line-height))
-               0)))))
+    (if (<= 27 emacs-major-version)
+	(+ y (window-tab-line-height) (window-header-line-height))
```
hiroyuki-komatsu: Please use space characters only, but not tab characters.

src/unix/emacs/mozc.el
```diff
@@ -652,15 +652,19 @@ However, on Emacs 24, y offset returned by `posn-at-point' is relative to
...
 
 This function fixes the asymmetry between them on Emacs 24 and later versions.
 This hack could be moved to mozc-posn-at-x-y in a future version."
   (let ((y (cdr (posn-x-y position))))
-    (if (or (null header-line-format) (<= emacs-major-version 23))
-        y
-      (+ y (or mozc-cached-header-line-height
-               (setq mozc-cached-header-line-height (mozc-header-line-height))
-               0)))))
+    (if (<= 27 emacs-major-version)
```
hiroyuki-komatsu: How about using `(fboundp #'window-tab-line-height)` instead of checking the version?

hiroyuki-komatsu: How about using `cond` instead of `if`?

hiroyuki-komatsu (commented):
Hi kkatsuyuki, thank you for your PR!

I've left some comments. Please take a look.

Also, could you please update the PR title to clearly describe the changes? For example: "Emacs: Fix candidate overlay collapse when tab-line is enabled"

Thank you!

src/unix/emacs/mozc.el
kkatsuyuki: Thanks, I did it.

src/unix/emacs/mozc.el
kkatsuyuki: I've done it.

kkatsuyuki: Done.

src/unix/emacs/mozc.el
```diff
@@ -658,13 +658,13 @@ line on Emacs 24.
...
-	(+ y (or mozc-cached-header-line-height
-		 (setq mozc-cached-header-line-height (mozc-header-line-height))
-		 0))))))
+    (cond ((fboundp #'window-tab-line-height)
+           (+ y (window-tab-line-height) (window-header-line-height)))
+          ((or (null header-line-format) (<= emacs-major-version 23))
+           y)
+          (t (+ y (or mozc-cached-header-line-height
+                      (setq mozc-cached-header-line-height
+                            (mozc-header-line-height)) 0))))))
```
hiroyuki-komatsu: How about keeping a line break before `0`? This would align the three values for `or` at the same indent level. (A line break after `t` is also an option).

```lisp
          (t
           (+ y (or mozc-cached-header-line-height
                    (setq mozc-cached-header-line-height
                            (mozc-header-line-height))
                    0))))))
```

If the emacs built-in formatter does not allow this indentation, please let me know. In that case  let's keep the current style.

kkatsuyuki: Done.

hiroyuki-komatsu:
Hi kkatsuyuki,

We have merged your PR.

PR #1508, opened 2026-05-10
Define `NO_SHLWAPI_STRFCNS` to not define `StrCat`

yukawa:
## Description
This is a mechanical cleanup to avoid defining `StrCat` in `shlwapi.h`, which often requires us to `undef StrCat` before using `absl::StrCat`.

There must be no functional change in the final executables.

## Issue IDs
N/A

 - OS: All
 - Steps:

### DIFF
--- a/src/BUILD.bazel
+++ b/src/BUILD.bazel
@@ -222,6 +222,7 @@ cc_library(
             "_ATL_NO_HOSTING",
             "ID_TRACE_LEVEL=1",
             "_MIDL_USE_GUIDDEF_",
+            "NO_SHLWAPI_STRFCNS",
             "PSAPI_VERSION=2",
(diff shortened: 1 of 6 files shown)


src/converter/nbest_generator.h
```diff
@@ -50,8 +50,6 @@
 #include "prediction/suggestion_filter.h"
 #include "request/conversion_request.h"
 
-#undef STRICT  // minwindef.h has the definition.
```
hiroyuki-komatsu, line 53: I think this line shuold be kept.

yukawa: Ah, good catch. Reverted to the original steate.

yukawa:
Also rebased onto `master`.

PR #1512, opened 2026-05-12
Add constinit global string holder classes

yukawa:
## Description
These are utility classes for global strings whose value is known only at runtime. They are motivated by the on-going effort to remove `Singleton<T>`, which has surfaced several places in system_util.cc (directory caches, user-SID lookup, program-invocation-name holder) where the singleton wraps a string lookup that must run **after** the surrounding modules are ready.

A function-local `static std::string` would be the obvious replacement, but reintroduces two issues that `Singleton<T>` callers have learnt to avoid:

  * On Windows, function-local static initialization is not safe to invoke from `DllMain` because the compiler-generated guard can take locks that interact badly with the Loader Lock.
  * The resulting destructor runs at process exit in unspecified order relative to other globals.

Both classes address these keeping all foreign code outside the publish lock and by being trivially destructible (unless
`MOZC_NO_ATOMIC_FLAG_WAIT` is specified).

`ConstInitImmutableString` is the publish-once, read-many variant. It takes an `IdempotentInitializer` function pointer that is invoked lock-free on the first `GetOrInit()`, so the class is safe to use from `DllMain` even when the initializer itself acquires the Loader Lock. Subsequent reads return a `string_view` over stable, NUL-terminated storage on a lock-free fast path:

```cpp
#include "base/strings/const_init_immutable_string.h"

constinit mozc::ConstInitImmutableString<256> g_program_files_x86(
    []() -> std::string { return ComputeProgramFilesX86Path(); });

absl::string_view path = g_program_files_x86.GetOrInit();
```

`ConstInitMutableString` is the set-many, snapshot-read variant. `Set()` atomically replaces the stored value; `Get()` returns a `std::string` snapshot under a lock-protected copy.

```cpp
#include "base/strings/const_init_mutable_string.h"

constinit mozc::ConstInitMutableString<256> g_invocation_name;

void Init(absl::string_view name) { g_invocation_name.Set(name); }
std::string GetName() {
  return g_invocation_name.Get();
}
```

For platforms where `std::atomic_flag::wait()` is disallowed in favor of `absl::Mutex`, `MOZC_NO_ATOMIC_FLAG_WAIT` macro is provided to switch the implementation to the `absl::Mutex`-based one. This guarantees that `std::atomic_flag::wait()` is used only for `client` build flavors in `mozc_select`.

## Issue IDs
N/A

 - OS: All
 - Steps:
   1. `bazelisk test //base/strings/...  --build_tests_only`

### DIFF
--- a/src/base/strings/BUILD.bazel
+++ b/src/base/strings/BUILD.bazel
@@ -76,6 +76,58 @@ mozc_cc_test(
 )
 
 mozc_cc_library(
+    name = "const_init_immutable_string",
+    hdrs = ["const_init_immutable_string.h"],
+    defines = mozc_select(
+        client = [],
+        default = ["MOZC_NO_ATOMIC_FLAG_WAIT"],
+    ),
+    deps = ["//base/strings/internal:const_init_string_helpers"] + mozc_select(
+        client = [],
+        default = [
+            "@com_google_absl//absl/base:core_headers",
+            "@com_google_absl//absl/synchronization",
+        ],
+    ),
+)
+
+mozc_cc_test(
+    name = "const_init_immutable_string_test",
+    size = "small",
+    srcs = ["const_init_immutable_string_test.cc"],
+    deps = [
+        ":const_init_immutable_string",
+        "//testing:gunit_main",
+    ],
+)
+
+mozc_cc_library(
+    name = "const_init_mutable_string",
+    hdrs = ["const_init_mutable_string.h"],
+    defines = mozc_select(
+        client = [],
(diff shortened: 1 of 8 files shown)


yukawa:
Thank you for the detailed review. As we talked offline, I decided to abandon this commit for now as we have at least removed `Singleton<T>` from `system_util.cc` as follows:

 * https://github.com/google/mozc/pull/1523

Maybe we can revisit here later (e.g. creating Win32-specifc optimization somewhere in `src/base/win32/win_util.cc`), but for now let's close this for now to minimize the number of open pull requests.

PR #1515, opened 2026-05-13
Remove `Singleton` from `ipc_path_manager.cc`

yukawa:
## Description
As part of our ongoing efforts to remove the dependency on `Singleton`, this commit rewrites `ipc/ipc_path_manager.cc` with `absl::NoDestructor` rather than `Singleton`.

## Issue IDs
N/A

 - OS: ALl
 - Steps:
   1. Confirm you can still type with Mozc


hiroyuki-komatsu:
We have merged your PRs.

Issue #1518, opened 2026-05-29
`//engine:data_loader_test` is flaky on `macos-15`


yukawa:
## Description
This really isn't new, but we have seen a random failure of `//engine:data_loader_test` only in `macos-15` GitHub Actions runner for a long time.

https://github.com/google/mozc/actions/runs/26625129921/job/78459927628
```
//engine:data_loader_test                                                FAILED in 1.0s
  /Users/runner/Library/Caches/bazel/_bazel_runner/70589f76e9298332e2e2764a78cc063e/execroot/_main/bazel-out/darwin_arm64-dbg/testlogs/engine/data_loader_test/test.log
```

## Steps to reproduce
Any reliable way to reproduce the issue is still unknown. So far it seems to be happening only on GitHub Actions runner for macOS. 

## Expected behavior
`//engine:data_loader_test` always passes.

## Actual behavior
`//engine:data_loader_test` sometimes fails.

## Version or commit-id
b30921c7e339bea70aba950d8cdec6299f550e6a

## Environment
 - OS: `macos-15` GitHub Actions runner

yukawa:
Here is my analysis report from the actual test failure reproduced in GitHub Actions.

## Root cause
`DataLoaderTest.WaitHighPriorityDataTimeoutTest` assumes the loader thread starts within ~100 ms.

Reproduced on macOS GitHub Actions runners (e.g. two of five shards failed in one push). The failing target is `//engine:data_loader_test`, and it always fails at the **same two assertions**:

```
engine/data_loader_test.cc:276: Failure   // wrong priority delivered to callback
engine/data_loader_test.cc:293: Failure   // callback_called != 2
```
https://github.com/google/mozc/blob/b18e5de65659cfa5a4f92066bcbb6bfa4d94dfe5/src/engine/data_loader_test.cc#L261-L294

Both failures stem from a **single timing race**, not two independent bugs.

### How the loader is supposed to behave

`DataLoader::StartReloadLoop` pops the highest-priority pending request, and *if no high-priority data has been registered yet* it waits up to `kTimeout = 100 ms` for a high-priority notification before committing to build the current request:

https://github.com/google/mozc/blob/b18e5de65659cfa5a4f92066bcbb6bfa4d94dfe5/src/engine/data_loader.cc#L197-L205

A request with `priority() <= kHighPriority (10)` is what fires that notification.
[...]

PR #1527, opened 2026-06-13
Upload failing test logs in GitHub Actions

yukawa:
## Description
This follows up on the flaky test failure (#1518), which we have recently fixed (#1522).

One thing that had contributed to the long time to fix the issue was that the test failure log files had not been uploaded as artifacts in GitHub Actions, which made it hard to investigate the root cause of the issue. To avoid similar situations in the future, let's make it possible to upload the log files as artifacts when there is any test failure in GitHub Actions.

To do so, this commit adds a new Python script `collect_test_logs.py` that collects failing test logs into the given directory so that the directory can be uploaded as a single artifact.

The core idea here is to rely on the [Build Event Protocol](https://bazel.build/remote/bep) (BEP) to retrieve the insight into a Bazel invocation. Unlike convenient symlinks such as `bazel-testlogs`, which for example cannot abstract away the complexity behind platform transitions while running all the tests in a single bazel invocation, the BEP gives the complete picture of the test execution including the absolute paths to the test log files. Let's use this opportunity to demonstrate how Bazel is powerful and flexible with the BEP.

This commit also adds `--build_tests_only` to the Linux test invocation so that it is consistent with the macOS and Windows workflows, which already pass that flag.

No artifact will be uploaded if there is no test failure.

## Issue IDs
 * https://github.com/google/mozc/issues/1518

 - OS: All
 - Steps:
   1. Explicitly trigger test failure in GitHub Actions
   2. Confirm that test failure logs were uploaded

### DIFF
--- a/src/build_tools/collect_test_logs.py
+++ b/src/build_tools/collect_test_logs.py
@@ -1,0 +1,141 @@
+# Copyright 2010-2021, Google Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
+#     * Redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above
+# copyright notice, this list of conditions and the following disclaimer
+# in the documentation and/or other materials provided with the
+# distribution.
+#     * Neither the name of Google Inc. nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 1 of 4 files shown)


src/build_tools/collect_test_logs.py
```diff
@@ -0,0 +1,122 @@
...
+that symlink is not produced reliably when the test command triggers platform
+transitions, and its location depends on the compilation mode. The URIs in the
+BEP are always correct for the run that produced them, so this works the same
+on Linux, macOS and Windows.
+"""
+
+import argparse
+import json
+import shutil
+from pathlib import Path
```
hiroyuki-komatsu: Please do not import a member of a module.

https://google.github.io/styleguide/pyguide.html#22-imports

```
import pathlib
...
def _relative_dest(src: pathlib.Path, label: str) -> pathlib.Path
```

src/build_tools/collect_test_logs.py
```diff
@@ -0,0 +1,122 @@
...
+from urllib.parse import urlparse
```
hiroyuki-komatsu: Please do not import a member of a module.

```
from urllib import parse
from urllib import request
```

src/build_tools/collect_test_logs.py
```diff
@@ -0,0 +1,122 @@
...
+from urllib.request import url2pathname
+
+
+def _relative_dest(src: Path, label: str) -> Path:
+  """Computes the destination path under the output dir for a source file.
```
hiroyuki-komatsu, line 59: Please add Args and Returns sections to the docstring.

hiroyuki-komatsu (commented):
Thank you for your PR!

src/build_tools/collect_test_logs.py
yukawa: Done in 3c415e384dd66a19e6b63e49169da587a7c47eb3.

src/build_tools/collect_test_logs.py
yukawa: Done in 3c415e384dd66a19e6b63e49169da587a7c47eb3.

src/build_tools/collect_test_logs.py
yukawa: Done in 3c415e384dd66a19e6b63e49169da587a7c47eb3.

hiroyuki-komatsu:
We have merged your PRs.
Thank you for your contributions!

PR #1529, opened 2026-06-13
Embed Win32 manifest into tests and build tools

### LINKED ISSUE #1528: Win32 executables built for tests and build tools do not have Win32 manifest
## Description
This is something we overlooked when migrating from GYP to Bazel, but currently intermediate build tools and unit test executables built on Windows do not have Win32 manifests.

## Steps to reproduce
1. `bazelisk build //base:bits_test -c dbg`

## Expected behavior
`bazel-bin/base/bits_test.exe` does have some Win32 manifests.
When built with GYP, it at least had the following entry
```xml
  <trustInfo>
    <security>
      <requestedPrivileges>
        <requestedExecutionLevel level="asInvoker" uiAccess="false"/>
      </requestedPrivileges>
    </security>
  </trustInfo>
```
Ideally it should be similar enough to production binaries as otherwise tests and production binaries would be running in different execution context.
https://github.com/google/mozc/blob/347a65f1cc42155e93e5ba2ac13b9c08f4102302/src/server/mozc_server.exe.manifest#L11-L22

## Actual behavior
`bazel-bin/base/bits_test.exe` does not have any Win32 manifest.

## Version or commit-id
347a65f1cc42155e93e5ba2ac13b9c08f4102302

## Environment
 - OS: Windows 11 25H2

## Additional context
Certain APIs change their behavior depending on the process' Operating System Context. This is particularly important for Mozc because `mozc_tip32.dll` and `mozc_tip64.dll` cannot choose in which Operating System Context the code would run. For instance,  we still need the following workaround in case `mozc_tip64.dll` is running in a process under Windows Vista context.
https://github.com/google/mozc/blob/347a65f1cc42155e93e5ba2ac13b9c08f4102302/src/ipc/win32_ipc.cc#L274-L306

yukawa:
## Description
Currently unit tests and intermediate executables used during the build have no Win32 manifest at all, while production binaries specify the supported OS versions in their manifests. This means that tests run under a different OS compatibility context from production binaries, and executables without any manifest are also subject to the UAC installer-detection heuristic.

This is partly a regression from the GYP to Bazel migration. The GYP build linked every executable, including unit tests, with the `link_embed` rule (`gyp-win-tool link-with-manifests <arch> True ...`), which embedded a default manifest declaring
`requestedExecutionLevel="asInvoker"` into every executable. Note that the supported OS versions section was specified only for production binaries even in the GYP build though.

To fix the discrepancy, embed a default manifest, which consists of the compatibility section shared by all the production manifests, into all the executables built with `mozc_cc_test` and `mozc_cc_binary` on Windows.

The default manifest itself only carries the compatibility section and does not declare a `trustInfo` section, exactly like the production manifests. The execution level is still embedded: whenever a manifest is embedded into an executable (`/MANIFEST:EMBED`), the linker adds its default `trustInfo` section with `requestedExecutionLevel="asInvoker"`. As a result every test and intermediate executable now ends up with the same `asInvoker` level that both the production binaries and the old GYP build embed, restoring parity without having to spell it out in the manifest source.

Production binaries opt out and keep controlling their own manifests through `mozc_win32_resource_from_template`. Shared libraries are also skipped as an application manifest is not applicable to them.

Closes #1528.

## Issue IDs
 * https://github.com/google/mozc/issues/1528

 - OS: Windows 25H2
 - Steps:
   1. `bazelisk build //base:bits_test -c dbg`
   2. Confirm `bazel-bin/base/bits_test.exe` embeds Win32 manifest in its resource section.

### DIFF
--- a/src/bazel/win32/BUILD.bazel
+++ b/src/bazel/win32/BUILD.bazel
@@ -32,10 +32,22 @@
 #  - Windows SDK library link options.
 #  - (that's it for now)
 
+load("//:build_defs.bzl", "MOZC_TAGS", "mozc_win32_resource")
 load("//bazel:stubs.bzl", "bzl_library")
 load("//bazel/win32:internal.bzl", "mozc_win32_lib")
 
 package(default_visibility = ["//:__subpackages__"])
+
+# A linker input to embed the default Win32 manifest settings (e.g. the
+# supported OS versions) that all the production binaries specify. This is
+# added to all the executables built with mozc_cc_test and mozc_cc_binary
+# unless opted out explicitly.
+mozc_win32_resource(
+    name = "win32_default_manifest",
+    manifests = ["win32_default.manifest"],
+    tags = MOZC_TAGS.WIN_ONLY,
+    target_compatible_with = ["@platforms//os:windows"],
+)
 
 bzl_library(
     name = "build_defs_bzl",
--- a/src/build_defs.bzl
+++ b/src/build_defs.bzl
@@ -74,6 +74,13 @@ def _copts_unsigned_char():
         "//conditions:default": ["-funsigned-char"],
     })
 
+def _win32_default_manifest_deps():
+    """Returns deps to embed the default Win32 manifest settings on Windows."""
+    return select({
+        "@platforms//os:windows": ["//bazel/win32:win32_default_manifest"],
+        "//conditions:default": [],
+    })
+
 def _update_visibility(visibility = None):
     """
     Returns updated visibility. This is temporarily used for the code location migration.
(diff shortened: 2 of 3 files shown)


src/bazel/win32/BUILD.bazel
```diff
@@ -32,11 +32,24 @@
 #  - Windows SDK library link options.
 #  - (that's it for now)
 
+load("@windows_sdk//:windows_sdk_rules.bzl", "windows_resource")
```
hiroyuki-komatsu: I have introduced `mozc_win32_resource`. Please use it.

`mozc_win32_resource` is an alias of `windows_resource` for this OSS build and another rule for the internal build.

src/build_defs.bzl
```diff
@@ -100,12 +107,23 @@ register_extension_info(
...
+      deps: deps for cc_binary.  //:macro is added.
+      copts: copts for cc_binary.
+      linkopts: linkopts for cc_binary.
+      enable_win32_default_manifest: optional. If True (default), embed the
+          default Win32 manifest settings on Windows. Ignored for shared
+          libraries, where an application manifest is not applicable.
+      **kwargs: other args for cc_binary.
     """
+    win32_default_manifest_deps = []
+    if enable_win32_default_manifest and not kwargs.get("linkshared"):
```
hiroyuki-komatsu, line 128: Please add a comment why `win32_default_manifest_deps` is not added when `linkshared` is `True`.

src/bazel/win32/BUILD.bazel
yukawa: Done.

src/build_defs.bzl
yukawa: Done.

PR #1535, opened 2026-06-27
Make clang-cl toolchains host-agnostic

yukawa:
## Description
As part of our on-going effort towards making it possible to build Mozc on Windows Arm64 machines (#1296), this commit further updates our local patches for rules_cc to make its clang-cl toolchains host-agnostic.

  * Add an explicit `--target=` to the x64 and x86 configs. Without it the Arm64-native clang-cl would fall back to its host default (aarch64) and miscompile the x64/x86 targets. The arm64 config already did this.
  * Drop the `@platforms//cpu:x86_64` `exec_compatible_with` pin from the x64 and x86 toolchains, matching what the arm64 toolchain already does, so they resolve on any Windows clang-cl host.

Above changes are relevant only when building Mozc on Windows Arm64 machines. There should be no observable changes in existing build pipelines running on Windows x64 hosts.

## Issue IDs


hiroyuki-komatsu:
We have merged your PRs.

Issue #1539, opened 2026-07-04
Consider enabling Segment Heap in Mozc executable for Windows


yukawa:
## Description
Visual Studio 2026 version 18.6 and later embeds Win32 manifests with Segment Heap enabled by default.

[Segment Heap support for C++ projects in Visual Studio](https://devblogs.microsoft.com/cppblog/segment-heap-support-for-c-projects-in-visual-studio/)
> Visual Studio 2026 version 18.6 makes it easier to take advantage of modern Windows memory management improvements. [Segment Heap](https://learn.microsoft.com/windows/win32/sbscs/application-manifests#heaptype) is a modern heap implementation in Windows that delivers stronger protection against common memory vulnerabilities, higher allocation throughput, lower memory fragmentation, better scalability across cores, and more predictable performance under load. Starting with this release, new C++ projects are now configured to use Segment Heap by default.

Assuming Segment Heap would deliver stronger protection against common memory vulnerabilities, higher allocation throughput, lower memory fragmentation, better scalability across cores, and more predictable performance under load as Microsoft has claimed, let's enable it for our executables such as `mozc_server.exe`.

To enable Segment Heap, we just need to update our Win32 manifest as follows:

https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#heaptype
> Overrides the default heap implementation for the [Win32 heap APIs](https://learn.microsoft.com/en-us/windows/win32/memory/heap-functions) to use.
>
> * The value **SegmentHeap** indicates that segment heap will be used. Segment heap is a modern heap implementation that will generally reduce your overall memory usage. This element is supported in Windows 10, version 2004 (build 19041) and later.
> * All other values are ignored.
>
> This element has no attributes.
> ```xml
> <assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
>  ...
>   <asmv3:application>
>     <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2020/WindowsSettings">
>       <heapType>SegmentHeap</heapType>
>     </asmv3:windowsSettings>
>   </asmv3:application>
>  ...
> </assembly>
> ```

Note that our code still runs in the legacy NT-heap process in the following cases:

 * The code is compiled into `mozc_tip64.dll`, and `mozc_tip64.dll` is loaded into NT-heap process.
[...]

PR #1540, opened 2026-07-05
Unify XML namespace style in Win32 manifest files

yukawa:
## Description
As a preparation for the upcoming change to opt-in Segment Heap (#1539) by adding a new entry to the Win32 manifest files, this commit unifies the XML namespace style.

With this commit, each element will be declared with its own default XML namespace instead of using namespace prefix aliases, following the style used in the current Microsoft manifest documentation. The prefix aliases were a workaround for a Manifest Tool bug that could corrupt the heap in sxs.dll on Windows XP SP2, which is no longer relevant. Also write the `dpiAwareness` element on a single line for consistency with the other settings.

No behavior change is intended.

## Issue IDs

 * https://github.com/google/mozc/issues/1539

 - OS: Windows 11 25H2
 - Steps:
   2. Enable Mozc to type something.
   3. Make sure that "Operating System Context" , "UAC Virtualization", and "DPI Awareness" shown in the task manager remain unchanged for those without this commit.

### DIFF
--- a/src/bazel/win32/win32_default.manifest
+++ b/src/bazel/win32/win32_default.manifest
@@ -1,24 +1,17 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
-  <!--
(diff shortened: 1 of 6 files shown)


hiroyuki-komatsu:
We have merged your PRs.

PR #1563, opened 2026-08-27
Guard FillIncognitoCandidateWords against empty incognito segments.

rootkiller6788:
## What

`EngineConverter::FillIncognitoCandidateWords()` accesses `incognito_segments_.conversion_segment(segment_index_)` without checking whether the container is non-empty.

## Root cause

When `fill_incognito_candidate_words` is set, `SuggestWithPreferences()` clears `incognito_segments_` and calls `StartPrediction()` in incognito mode. If that prediction fails (e.g. zero-query suggestion where the only candidates come from user history, which is disabled in incognito mode), the code falls through (see the existing `TODO(noriyukit): Check if fall through here is ok.`) and `incognito_segments_` stays empty. The state is still set to `SUGGESTION`, so a subsequent `FillOutput()` unconditionally calls `FillIncognitoCandidateWords()` (whenever the request flag is set), which then dereferences `conversion_segment(0)` on an empty `Segments` container. `Segments::conversion_segment()` is `*segments_[i + history_segments_size()]` with no bounds check, so this is an out-of-bounds read.

## Fix

Add the same size guard that the sibling method `FillAllCandidateWords()` already has:

```cpp
if (segment_index_ >= incognito_segments_.conversion_segments_size()) {
  LOG(WARNING) << "Invalid segment_index_: " << segment_index_
               << ", incognito_segments_size: "
               << incognito_segments_.conversion_segments_size();
  return;
}
```

## Test

Added `SuggestFillIncognitoCandidateWordsEmptySegments`, which mocks a successful normal suggestion followed by a failing incognito prediction and asserts that `FillOutput()` produces no incognito candidate words instead of crashing.

### DIFF
--- a/src/engine/engine_converter_test.cc
+++ b/src/engine/engine_converter_test.cc
@@ -2314,6 +2314,47 @@ TEST_F(EngineConverterTest, SuggestFillIncognitoCandidateWords) {
   }
 }
 
+TEST_F(EngineConverterTest, SuggestFillIncognitoCandidateWordsEmptySegments) {
+  Segments segments;
+  {  // Initialize mock segments for suggestion
+    Segment* segment = segments.add_segment();
+    converter::Candidate* candidate;
+    segment->set_key(kChars_Mo);
+    candidate = segment->add_candidate();
+    candidate->value = kChars_Mozukusu;
+    candidate->content_key = kChars_Mozukusu;
+    candidate = segment->add_candidate();
+    candidate->value = kChars_Momonga;
+    candidate->content_key = kChars_Momonga;
+  }
+  composer_->InsertCharacterPreedit(kChars_Mo);
+
+  // A matcher to test if the given conversion request sets incognito_mode().
+  constexpr auto IsIncognitoConversionRequest = [](bool is_incognito) {
+    return Property(&ConversionRequest::incognito_mode, Eq(is_incognito));
+  };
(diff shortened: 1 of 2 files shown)


src/engine/engine_converter_test.cc
```diff
@@ -2324,6 +2314,47 @@ TEST_F(EngineConverterTest, SuggestFillIncognitoCandidateWords) {
...
+              StartPrediction(IsIncognitoConversionRequest(false), _))
+      .WillOnce(DoAll(SetArgPointee<1>(segments), Return(true)));
+  EXPECT_CALL(*mock_converter,
+              StartPrediction(IsIncognitoConversionRequest(true), _))
+      .WillOnce(Return(false));
+
+  EXPECT_TRUE(converter.Suggest(*composer_, Context::default_instance()));
+  commands::Output output;
+  converter.FillOutput(*composer_, &output);
+  EXPECT_FALSE(output.has_incognito_candidate_words());
```
hiroyuki-komatsu, line 2355: This `EXPECT_FALSE` check fails in my environment.

To address it, `segment_index_ >= incognito_segments_.conversion_segments_size()` should be called before `mutable_incognito_candidate_words()`.

hiroyuki-komatsu (commented):
Thank you for your PR.

However, the test fails in my environment. Could you please double check it in your environment?

PR #1568, opened 2026-09-16
Optimize `strings::CharsLen` with `std::popcount`

yukawa:
## Description
Replace the per-character loop that advances by `OneCharLen` with a count of the bytes that are not UTF-8 continuation bytes, processed 8 bytes at a time with bit operations and `std::popcount`. Both approaches return the same count for valid UTF-8 strings, which `CharsLen` already requires.

The public interface keeps the `absl::string_view` overload unchanged. The iterator-pair overload now requires contiguous iterators instead of input iterators so that it can forward to the `string_view` overload; all existing callers pass `string_view` or its iterators. `Util::CharsLen`, which had its own copy of the loop, now delegates to `strings::CharsLen` so that its many callers benefit as well.

Results of `//base/strings:unicode_benchmark_test` (14e69ba8411b49712714c4634eb4612dacaeabcb) on Windows x64 (Intel Core i9-13900KF, clang-cl 20.1.1, -c opt, medians of 10 interleaved runs), before -> after:

```
  BM_CharsLen/ascii_long     16681 ns -> 542 ns (30.8x)
  BM_CharsLen/greek_long      8306 ns -> 539 ns (15.4x)
  BM_CharsLen/mixed_long      6829 ns -> 545 ns (12.5x)
  BM_CharsLen/kanji_long      5536 ns -> 538 ns (10.3x)
  BM_CharsLen/emoji_long      4178 ns -> 545 ns ( 7.7x)
  BM_CharsLen/ascii_short     58.7 ns -> 5.5 ns (10.7x)
  BM_CharsLen/mixed_short     89.6 ns -> 12.0 ns ( 7.5x)
  BM_CharsLen/greek_short     15.7 ns -> 4.8 ns ( 3.3x)
  BM_CharsLen/kanji_short     8.09 ns -> 4.8 ns ( 1.7x)
  BM_CharsLen/emoji_short     2.34 ns -> 4.8 ns ( 0.5x)
```

Throughput is now data-independent at about 14.4 GiB/s for long inputs, and short inputs below the vectorization threshold cost about 5 ns regardless of content.

The generated code explains the numbers. The old loop compiles to six instructions per character but has a loop-carried dependency through two chained loads: the byte load feeds the length-table load, which feeds the pointer increment for the next iteration, so the loop is bound by load latency and can be neither unrolled nor vectorized. The new loop is auto-vectorized by clang into an SSE2 loop that processes 32 bytes per iteration, computing the byte-wise population count with shift/mask operations and accumulating it with `psadbw`. This vectorization is also why the loop is written with `absl::Span::remove_prefix`: the same computation with explicit index arithmetic (i + 8 <= n) stays scalar at one 8-byte word per iteration, with `std::popcount` expanded to the multiply-based fallback because the x86-64 baseline target has no `POPCNT` instruction, and measured about 35% slower (9.5 GiB/s vs 14.4 GiB/s).

Results on Apple M1 (arm64 with libc++, -c opt, medians of 10 interleaved runs), before -> after:

```
  BM_CharsLen/ascii_long     23247 ns -> 255 ns (91.3x)
  BM_CharsLen/greek_long     11583 ns -> 251 ns (46.2x)
  BM_CharsLen/mixed_long      9471 ns -> 254 ns (37.3x)
  BM_CharsLen/kanji_long      7726 ns -> 251 ns (30.8x)
  BM_CharsLen/emoji_long      5795 ns -> 253 ns (22.9x)
  BM_CharsLen/ascii_short     65.5 ns -> 6.3 ns (10.4x)
  BM_CharsLen/mixed_short    110.6 ns -> 8.0 ns (13.9x)
  BM_CharsLen/greek_short     25.8 ns -> 3.8 ns ( 6.7x)
  BM_CharsLen/kanji_short     13.6 ns -> 3.8 ns ( 3.6x)
  BM_CharsLen/emoji_short      3.6 ns -> 5.4 ns ( 0.7x)
```

Throughput for long inputs is again data-independent, at about 30.3 GiB/s, and the disassembly shows the same two mechanisms as on x86-64. The old loop compiles to the same chained
```
  byte-load -> table-load -> pointer-add
```
sequence, whose two dependent L1 loads cost about 2.8 ns per character on M1 (the measured `ascii_long` time matches this exactly). The new loop is auto-vectorized by clang into a NEON loop that processes 64 bytes per iteration (bic/and to build the continuation-byte mask, `cnt`/`udot`/`uadalp` to accumulate), which is why the long-input speedups exceed the x86-64 ones, whose SSE2 baseline lacks `POPCNT`.

The one regression, on both platforms, is `emoji_short` (28 bytes), the old code's best case: seven 4-byte characters cost only seven iterations of the old loop, while the new code stays below the vectorization threshold and processes three 8-byte words plus a 4-byte byte-wise tail, paying for `std::popcount` on each word (the multiply-based fallback on x86-64 without `POPCNT`; a round trip through a NEON register on arm64, which has no scalar `popcount` instruction). The absolute cost is about 2 ns on a 28-byte string, and strings that are both this short and made up purely of 4-byte characters are rare in Mozc.

## Issue IDs
N/A

 - OS: All
 - Steps:
   1. `bazelisk run -c opt //base/strings:unicode_benchmark_test -- --benchmark_repetitions=9 --benchmark_min_warmup_time=0.5 --benchmark_enable_random_interleaving=true --benchmark_report_aggregates_only=true`


src/base/strings/unicode.cc
```diff
@@ -29,16 +29,40 @@
...
 namespace mozc {
 namespace strings {
 
+size_t CharsLen(const absl::string_view sv) {
+  // Count the bytes that are not UTF-8 continuation bytes (0b10xxxxxx), i.e.
+  // the leading byte of each character, processing eight bytes at a time with
+  // bit operations.
+  constexpr uint64_t kHighBits = 0x8080808080808080;
+  constexpr size_t kBlockSize = sizeof(uint64_t);
+  absl::Span<const char> chars = absl::MakeConstSpan(sv.data(), sv.size());
```
hiroyuki-komatsu, line 52: (non-const) `sv` can be used directly instead of making `absl::Span`.

src/base/strings/unicode.cc
```diff
@@ -29,16 +29,40 @@
...
+  size_t continuations = 0;
+  while (chars.size() >= kBlockSize) {
+    uint64_t w;
+    std::memcpy(&w, chars.data(), kBlockSize);
```
hiroyuki-komatsu, line 56: `LoadUnaligned<uint64_t>` is available in `mozc/src/base/bits.h`. 

```
const uint64_t w = LoadUnaligned<uint64_t>(sv.data());
```

`absl/types/span.h` can be removed.

src/base/strings/unicode.cc
```diff
@@ -29,16 +29,40 @@
...
+    continuations += std::popcount(w & ~(w << 1) & kHighBits);
```
hiroyuki-komatsu, line 57: Please add a comment about this technique (i.e. count bytes that only start with `0b10xx`).

src/base/strings/unicode.cc
```diff
@@ -29,16 +29,40 @@
...
+    chars.remove_prefix(kBlockSize);
+  }
+  for (const char c : chars) {
+    continuations += (static_cast<uint8_t>(c) & 0xc0) == 0x80;
```
hiroyuki-komatsu, line 61: `continuations += static_cast<int8_t>(c) < -64;` is more efficient?

| bit | value | 
| --- | --- |
| 0x | 0 ... 127 |
| 11 | -64 ... -1 |
| 10 | -128 ... -65 |

src/base/strings/unicode.h
```diff
@@ -82,14 +82,18 @@ bool IsValidUtf8(absl::string_view sv);
...
+// the leading byte of each character and doesn't check if it's well-formed.
 // Complexity: linear
+size_t CharsLen(absl::string_view sv);
 template <typename InputIterator>
-  requires std::input_iterator<InputIterator>
-size_t CharsLen(InputIterator first, InputIterator last);
-inline size_t CharsLen(const absl::string_view sv) {
-  return CharsLen(sv.begin(), sv.end());
+  requires std::contiguous_iterator<InputIterator>
+size_t CharsLen(const InputIterator first, const InputIterator last) {
```
hiroyuki-komatsu, line 91: Is it possible to define CharLen as follows?

```
template <std::contiguous_iterator InputIterator>
  requires std::same_as<std::iter_value_t<InputIterator>, char>
size_t CharsLen(const InputIterator first, const InputIterator last) {
  return CharsLen(absl::string_view(first, last));
}
```

